PR #8, opened 2015-07-15
Add govcms_dlm module

paulkilla:
Add govcms_dlm module so that emails sent through drupal_mail send with DLM/Classification when enabled


drupal-org.make
```diff
@@ -59,6 +59,7 @@ projects[google_analytics_reports][patch][] = "https://www.drupal.org/files/issu
 projects[google_analytics_reports][version] = "1.3"
 projects[google_chart_tools][patch][] = "https://www.drupal.org/files/issues/google-chart-tools-ajax-1613258-11-7.x-1.4.patch"
 projects[google_chart_tools][version] = "1.4"
+projects[govcms_dlm][version] = "7.x-1.0"
```
seanhamlin: should be "1.0"

PR #10, opened 2015-07-15
Replace govCMS-Core references with govCMS.

stooit:


seanhamlin:
This is actively breaking the build, and helps to fix https://www.drupal.org/node/2532620

PR #14, opened 2015-07-16
Include chosen js library in build profile, enable on fields, add behat test.

stooit:


typhonius:
@srowlands can we get travis to print where applicable screenshots can be accessed from - perhaps using the `after_failure` directive? I downloaded this patch locally and couldn't replicate the failure so it'd be good to work out why travis has issues.

stooit:
Artifact upload is not supported on the `continuous-integration/travis-ci/pr` test unfortunately so it doesn't help us out there. That build looks to be ignoring the change to drupal-org.make (you can see in the log it never attempts to download the Chosen library, so it is a valid failure).

We have seen this behaviour on the `continuous-integration/travis-ci/pr` test on other PRs as well, would be good to understand why it differs.

typhonius:
Oh ok... so while the merged branch seems to work just fine (ie after everything is merged and added to the make files), the one where we're working on a different branch (ie branch is not master) is broken.

This looks related to the `make:local` `make:remote` differences we make in build.xml. cc/ @gollyg

seanhamlin:
Upon discussions with @srowlands it appears this is an issue with TravisCI, I think that should be a follow up PR, this one is fine if you run the tests locally.

PR #25, opened 2015-07-20
Make behat tests phantomjs compatible.

stooit:
Using @javascript means tests will run via headless phantomjs, so we can capture screenshots of failures as artifacts and auto-upload to S3.

The S3 bucket will be publicly accessible and if a failure occurs a link to the fail screenshot capture will display in Travis Logs (demo PR incoming shortly).

### DIFF
--- a/build/tests/behat/bootstrap/FeatureContext.php
+++ b/build/tests/behat/bootstrap/FeatureContext.php
@@ -123,14 +123,38 @@ class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext
   }
 
   /**
+   * Sets an id for the first iframe situated in the element specified by id.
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -58,7 +58,7 @@ before_script:
   - mysql -e 'CREATE DATABASE travis_ci_govcms_drupal;'
 
   # Build the codebase
-  - phing -f build/phing/build.xml build
+  - phing -f build/phing/build.xml build:local
```
typhonius, line 61: Can we remove this from this PR and let that be the job of #27

stooit, line 61: Definitely, will re-roll when/if #27 is merged

fiasco:
This is awesome, nice one Stu!

PR #27, opened 2015-07-20
Use build:local in Travis to avoid false failures

stooit:


typhonius:
Do we need to further document the difference between `build:local` and `build:remote` and when you would use each in `README.md`

seanhamlin:
Let's file a follow up ticket to remove build:remote all together (and update the documentation).

typhonius:
I think there's still value in `build:remote` but perhaps only when the user specifically requests it. Ie, not ever in the default path.

seanhamlin:
@typhonius are you happy for that to be in a follow up PR?

typhonius:
Created #29 to deal with this :feet:

PR #28, opened 2015-07-21
Add in pathologic and configure by default relative settings

paulkilla:
Add in pathologic 3.0-beta2 to help with showing alias url's in source instead of node id's etc, and to help with environment migrations


seanhamlin:
Had a chat with Paul around this, and this is going to be added into features for govCMS. At the moment the text filter is added only once on install, and never tracked. We have added it to features, and added an update hook to enable pathologic. Everything seems sane from my perspective.

modules/features/govcms_editing/govcms_editing.info
```diff
@@ -3,6 +3,7 @@ core = 7.x
 package = govCMS core
 project = govcms_editing
 dependencies[] = ctools
+dependencies[] = defaultconfig
```
seanhamlin, line 6: @paulkilla can we remove this dependency and use strongarm?

paulkilla:
@seanhamlin I don't think so, it is using the same functionality as the govcms_install function on site install. If someone wants to attempt it, my feeling won't be hurt :-)

seanhamlin:
After chatting to Paul, this looks good to go, defaultconfig is already a dependency of govCMS, so this is not new.

PR #38, opened 2015-08-04
Add theme hook to allow files to be cached per revision

paulkilla:
By adding the timestamp query string to file entities


seanhamlin:
All looks fine here, merging in.

PR #39, opened 2015-08-06
add update hook to configure on existing sites

paulkilla:
This is to configure existing sites, as why wouldn't anyone want pathologic setup?


seanhamlin:
Will wait for TravisCI to pass, thanks @paulkilla !

PR #47, opened 2015-09-04
Remove conditional classes

MadeByMike:
I think you should strongly consider removing conditional classes all together with feature detection being a strong preference over browser targeting. See reasons here: http://www.impressivewebs.com/conditional-comments-classes-ie/

The only class name required by the stylesheets appears to be "lt-ie9". You could add a few others if you intended these to be hooks for further development so feel free to modify this PR but there is no situation where IE7Mobile or ie6 should need to be accounted for.

### DIFF
--- a/themes/govcms/govcms_zen/templates/html.tpl.php
+++ b/themes/govcms/govcms_zen/templates/html.tpl.php
@@ -7,11 +7,9 @@
  * @see https://drupal.org/node/1728208
  */
 ?><!DOCTYPE html>
-<!--[if IEMobile 7]><html class="iem7" <?php print $html_attributes; ?>><![endif]-->
-<!--[if lte IE 6]><html class="lt-ie9 lt-ie8 lt-ie7" <?php print $html_attributes; ?>><![endif]-->
-<!--[if (IE 7)&(!IEMobile)]><html class="lt-ie9 lt-ie8" <?php print $html_attributes; ?>><![endif]-->
-<!--[if IE 8]><html class="lt-ie9" <?php print $html_attributes; ?>><![endif]-->
-<!--[if (gte IE 9)|(gt IEMobile 7)]><!--><html <?php print $html_attributes . $rdf_namespaces; ?>><!--<![endif]-->
+
+<!--[if (lt IE 9) ]><html class="lt-ie9" <?php print $html_attributes; ?>> <![endif]-->
+<!--[if (gt IE 9)|!(IE)]><!--> <html <?php print $html_attributes; ?>> <!--<![endif]-->
 
 <head>
   <?php print $head; ?>


fiasco:
I like the idea but I wouldn't call myself an expert in this area. What happens to older browsers? As we all know, they still exist in government.

MadeByMike:
Hi Josh, generally these classes are added to target older browsers in CSS. For JavaScript modernizer or custom feature detection should be used. 

IE hacks (star hack, underscore hack or backslash hacks) can still be used in stylesheets if required; but if they're required I'd say there is a strong chance the team is approaching CSS the wrong way. A progressive enhancement approach should mean that most styles just work, even if they are presented in a slightly degraded format -such as no rounded corners in older versions of IE.

If you decide you still need conditional classes they should be a sensible set something like "ie7 ie8 ie9 ltie9 ltie10". I'd suggest even less is required and would encourage better theme development.

aleayr:
Thanks Mike, I don't have an objection to this, just wondering for more experienced Drupal-ers, seems the conditionals on the last item (gte IE 9)|(gt IEMobile 7) includes $rdf_namespaces, which this PR doesn't include.

Not required?

MadeByMike:
Hi Adam,

It's a bit odd that it was applying $rdf_namespaces only when none of the conditionals were met. So they were not being applied correctly anyway. 

Does Drupal even do rdf namespacing correctly for a HTML5 document?: http://www.w3.org/TR/html-rdfa/#preserving-namespaces-via-coercion-to-infoset 

Also I'm no expert on this, but is there any real situation where you would want to declare this? I've genuinely never seen this used outside of XML and XHTML.

fiasco:
We do need $rdf_namespaces kept in the template. I think the rdfa module needs to be on for this to work properly but we shouldn't take that option away at the theme level.

MadeByMike:
@fiasco can you confirm the use case for rdf namespaces? I'd think it's such an edge case that it shouldn't be in the default template. 

That said, I guess it won't do a lot of harm if the rdf module is not enabled. Be aware that it isn't currently implemented correctly anyway. Currently the rdf namespaces will only be output in ie10+ and not <ie10. 

To apply this correctly with conditional classes you'd need to add it to every conditional output. It would be something like this:

```
<!DOCTYPE html>
<!--[if lte 7]><html class="ie7 lt-ie8 lt-ie9 lt-ie10" <?php print $html_attributes; . $rdf_namespaces; ?>><![endif]-->
<!--[if IE 8]><html class="ie8 lt-ie9 lt-ie10" <?php print $html_attributes . $rdf_namespaces; ?>><![endif]-->
<!--[if IE 9]><html class="ie9 lt-ie10" <?php print $html_attributes . $rdf_namespaces; ?><![endif]-->
<!--[if gte IE 10]><html <?php print $html_attributes . $rdf_namespaces; ?>><![endif]-->
```

*Not Tested

fiasco:
AFAIK, RDF is not really for browser use but by other net crawlers wanting semantic information from your site. Schema.org and opengraph are similar concepts. So I don't think it matters that the namespace is provided to IE. From the research I conducted, I didn't see any evidence that any browser the RDFa namespace.

This namespace is [present in Drupal core](http://cgit.drupalcode.org/drupal/tree/modules/system/html.tpl.php?h=7.x) and so we should persist that in govCMS.

teamglenny:
Is this still a progressing thing?

fiasco:
@teamglenny yes it is. Awaiting @MadeByMike to preroll patch.

MadeByMike:
Am I? Pre roll patch? You want me to update the pull request based on discussions?

pandaskii:
It is an interesting post from Louis Lazaris http://www.impressivewebs.com/conditional-comments-classes-ie/ 

Paul Irish (who is from Google Chrome team) also has a post that lists full discussions and thoughts regarding "conditional stylesheets and CSS hacks" http://www.paulirish.com/2008/conditional-stylesheets-vs-css-hacks-answer-neither/

We are using "Zen" as a base theme for govCMS starter kit, and notes in the "Zen" starterkit.info.txt file

> ; Built-in conditional stylesheet support has been removed from Zen 7.x-5.x.
> ; Instead, Zen now adds conditional-comment-included classes to the html
> ; element, such as .lt-ie9, .lt-ie8, .lt-ie7. 

I would like to suggest that we follows the guidelines in our offical themes while Zen is used as a base theme. In another word, designers and developers can always alter them in custom theme template (html.tpl.php).

aleayr:
@jozhao So, is your suggestion to leave as is? We are looking at redoing the base theme currently, in which a move to Bootstrap over Zen is preferred, so all this could be resolved via that?

pandaskii:
@aleayr yes, I think we can leave it as at current stage

aleayr:
Closing, and will track further developments on theme outputs via #176.

PR #48, opened 2015-09-11
Added OpenAccess module

fiasco:
To provide an accessibility checker on wysiwyg fields. Help your content editors write and maintain content that complies with structural, language, and accessibility standards. Content editors can preview the available templates before selecting the one that suits their requirements.


murraywoodman:
I have had a look at the module and have a couple comments.
1. The performx library would be much cleaner if the "resources" were stored outside of it - either in a module or theme. As this is govCMS it would be very helpful to have this implementable by the theme. This would allow the WYSIWYG to be extended depending on the use case at hand.
2. There seems to be a big advertisement for Systemik in the pop up :)

We were excited to see this kind of functionality proposed to govCMS as we really want to extend the WYSIWYG for editors as well. If we are going to implement hook_wysiwyg_plugin() then we should definitely do it in a way so that it is accessible from the theme. This gives all implementors the chance to have some control over the WYSIWYG.

aleayr:
Hi Murray,

Sorry for the delayed response. Can I clarify your comments?

Firstly, are you supportive of this specific module being included in the distribution? I get the impression, that you're actually asking for an alternative approach that allows you as implementors the ability to hook in this type of functionality depending on the project/customer needs?

i.e. We make hook_wysiwyg_plugin/hook_wysiwyg_include_directory available, and then implementors can make additional changes to the WYSIWYG interface?

Am I understanding correctly? Look forward to your response.

murraywoodman:
Hi Adam. We were looking for similar functionality recently when we were building out the Govt Demo site. In particular we wanted to hook a few different templates into the WYSIWYG to deliver a few helpful tools to editors: a few different grids, etc. We were not able to achieve this from the theme layer and were therefore blocked from adding this functionality. So, I think that it would be a very helpful feature if the ability to add templates via the theme layer was added. I think that this is a natural fit and will open up a lot of extra flexibility where required.

With the current code, the templates are baked into the module/library and cannot be altered. The templates are very specific and may only address a small set of requirements. My suggestion was to introduce a more decoupled approach where new templates could be added as required to the theme, opening up the door for more possibilities.

We are walking a fine line here between themes and modules. Normally we would do this with a module. As templates are closely tied to an implementation in the theme layer I think a good case could be made for allowing this functionality in the theme. So yes, I would like to see this as alternative approach. It would still allow for the addition of PerformX OpenAccess, however, other templates could also be added if required.

aleayr:
Great, thanks for clarifying, that clears up the issues at play in my mind in any case. Let me run the idea past a few colleagues on this end, and if anyone else watching the queue wants to chime in with some thoughts or advice, that would be appreciated as well.

fiasco:
Would this be something that could be better managed at a base theme level?

teamglenny:
Can we pull this in as an initial first step and look to expand in the future?

fiasco:
The current pull request is simply that, just added the module. It doesn't turn it on by default, or configure it. I think the biggest issue is that this change favours a business (Systemik). So long as we're ok with that, I think we can merge this in.

teamglenny:
Will reach out to the module maintainers with the feedback.

teamglenny:
The folks at Systemik are going to update the module.

yangli0516:
The Systemik logo has been removed from the plugin. Please go to [here](http://ckeditor.com/addon/performx) to download the latest plugin (v1.0.1).

fiasco:
Hi @yangli0516 - this pull request references the Drupal module which will need to be updated to v1.0.1 also.

yangli0516:
Hi @fiasco The Drupal module stays the same as I only updated the external dependency to 1.0.1. I've updated the URL of the plugin in #49

pandaskii:
@aleayr  Should we close this pull request in favour of #49 ? 

[Add OpenAccess module, external library and initial setup](https://github.com/govCMS/govCMS/pull/49)

aleayr:
Yes, good point, thanks.

PR #53, opened 2015-09-30
Update Big Menu to 7.x-1.3

mstrelan:
Big Menu 7.x-1.3 fixes a bug [#1980810] causing the "Show row weights" link to be duplicated each time a menu parent is expanded. This release also fixes a couple of multilingual issues and includes a minor code cleanup.


seanhamlin:
This is a simple update and fixes an annoying issue, if the tests pass, lets get this one in.

typhonius:
Do we want to also create a behat test for the weights bug to flag regressions in this fix?

fiasco:
@typhonius that sounds like a fix that should be in the module itself no? I'm happy to pull this in. @mstrelan are you willing to write a behat test to catch any regressions of this?

PR #55, opened 2015-09-30
Update file_entity module to 2.0-beta2 and media module to 2.0-beta1

paulkilla:


drupal-org.make
```diff
@@ -72,9 +72,9 @@ projects[link][version] = "1.3"
 projects[login_destination][version] = "1.1"
 projects[login_security][version] = "1.9"
 projects[maintenance_mode_permissions][version] = "1.0"
-projects[media][patch][] = "https://drupal.org/files/media.code_.1701914-2.patch"
+;projects[media][patch][] = "https://drupal.org/files/media.code_.1701914-2.patch"
```
fiasco: Lets get rid of the line altogether. No need to keep it commented out. Thats what git is for ;)

paulkilla:
I've just rebased and squashed commits into 1

fiasco:
Looks good to me

aleayr:
Can we merge?

fiasco:
Yes, though there perhaps a small amount of risk we're not covering ourselves for here not really testing this. If we had some tests around this modules capability (or at least what we're expecting of it) then this would give us some more confidence to pull these types of things.

PR #56, opened 2015-10-02
As a content author I want to create a poll which allows multiple votes from 1 IP address (e.g. corporate networks typically have 1 public IP but many actual users).

fiasco:
This module changes IP address restriction to cookie based solution so that users behind a firewall will be able to vote.


paulkilla:
+1 this works fine

fiasco:
Thanks @paulkilla - I think if @srowlands can give his blessing as well, then we should move forward with this. Should it be standard? As in, should we essentially take poll out as an option (whitelisting)?

paulkilla:
I think for now we should just leave it in. Let the users decide which one they want to use, as by default neither will be enabled.

stooit:
This module does not help users already using core Polling functionality. It creates a separate entity type which requires it's own field configuration, display settings, view modes, etc.

If we are looking to add a new module to resolve existing issues this choice will create more work for existing users.

aleayr:
@srowlands Do you mean that this module doesn't seamlessly integrate and upgrade the functionality of the core Polling module (and thus they don't get the benefits without physically switching modules)? If so, then can you suggest an alternate option?

We have govCMS sites that need a solution to this limitation "yesterday", so your rapid advice is appreciated.

stooit:
That's right @aleayr -- this module does not enhance or interact with the core Poll module. If we are looking at introducing a module that 'fixes' an outstanding issue then this will not be the silver bullet. #57 **should** bolt-on to core Poll, but it has a different set of issues to overcome.

I'm not saying this isn't the right approach, simply highlighting the fact that this will not immediately resolve any issues without further effort.

The issue in itself is interesting. There are essentially two ways to allow anonymous interaction:
- IP based restrictions (current default)
- cookie based restrictions (easy to game, simply clear your local browser cookies and vote again)

One way or another there will be an issue to be found with anonymising Polls - you could argue it is 'broken' using either approach. Possibly the "best" solution would involve OAuth with a popular service provider (Twitter/FB/Google), but that would likely mean either looking to a third party solution or developing our own.

aleayr:
I think in the vast majority of cases, polls won't be looking to address scientific consensus. We're just after a solution which will allow people to quickly (and if they want to, anonymously) answer basic polling questions. If people want to game the system, we have to ask the question, do we care?

If so, what would be do about it? Probably close that poll and ignore the results. If not, then, lets just do this.

We can publish some info as to some of these perceived limitations of a cookie-based polling system so site builders/editors are aware, but I don't think we need to go overboard in stopping people voting.

fiasco:
As Pollim doesn't fit onto the existing poll module, I think we have to close it out here. Perhaps there is reason to add it later down the track, but its not a good fit now since we already rely on the Poll module in core. I'm going to close this issue out.

fiasco:
I've opened a new pull request (#59) for adding the predecessor of pollim, pollanon.

PR #57, opened 2015-10-02
As a content author I want to create a poll which allows multiple votes from 1 IP (#2 option)

fiasco:
The module has several advantages over core poll:
- Removes the 1 vote per IP restriction (still records the IP address so malicious votes can be purged)
- Voting is done via AJAX
- Poll voting is recorded locally in the browser with local storage (so you see the results if you already have voted)
- Poll results are returned via AJAX, and can be cached in varnish (with potentially a different TTL to the page)
- Leverages core poll's content creation ability
- Allows more than 1 option to be chosen, up to a configurable maximum, if more than one option is allowed, then checkboxes are presented rather than radio buttons


paulkilla:
So, the module lists "js" as a dependency:
From .info file
﻿﻿dependencies[] = js

This isn't in the govCMS distribution.

Edit: Looks like the [js project](https://www.drupal.org/project/js) itself has bigger issues e.g. compatability with admin menu, so might be better going with #56

seanhamlin:
@paulkilla the JS module dependency can be removed by rewriting the callback in hook_menu() - this should be trivial to implement. govCMS also does not use admin_menu and in any case there is an issue to fix the integration https://www.drupal.org/node/1981308

fiasco:
This module requires a lot of work. We need to release it, or actually bring it into govCMS itself (which is probably a better idea than the current pull request). Then we'll need to swap out ajax issues to reduce dependencies. This module also adds additional complexities that perhaps are not needed at the moment. I've opened #59 to at the pollanon module from Drupal contrib which appears to be a better fit though there are some maintainer concerned to be aware of.

fiasco:
Given the work to get this into shape and the fact that our requirements can be meet with an existing module (#59) - I'm going to close this one off.

PR #58, opened 2015-10-06
Add transliteration module

mstrelan:
Is there any reason not to use transliteration in govCMS?


fiasco:
No I don't believe there isn't. Should there be any enforced configuration or are we happy with the defaults?

mstrelan:
The defaults are good. The other option worth considering is enabling transliteration on url aliases generated by pathauto (variable name is pathauto_transliterate). As mentioned on the project page this is a matter of personal taste.

> Whether you use transliteration for URLs (when using Pathauto 2.x), however, is a matter of personal taste. For example, the russian Wikipedia does not transliterate, but uses full Unicode in URLs. On the other hand, as a user noted, links containing Unicode look quite ugly in e-mails sent from your site.

fiasco:
Can we set pathauto_transliterate to TRUE in the govcms install profile?

aleayr:
I don't think it's going to cause us issues to do this. Might run it past a couple of people on this end and update shortly.

fiasco:
This won't break existing sites since it doesn't retrospectively turn on. So I'll commit this. Thanks!

PR #59, opened 2015-10-06
As a content author I want to create a poll which allows multiple votes from 1 IP address

fiasco:
Added poll anon module to augment the core poll module.This module enables anonymous users to vote on polls without page cache troubles or IP address restrictions. It is designed to work on top of core Poll module. See https://www.drupal.org/project/pollanon for more information.

One issue with this module however is that it is essentially no longer maintained. While it does augment the core poll module which is what we're after, the maintainer believes the pollim module is the future and successor which we've already rejected in #56.


stooit:
This sounds like it addresses the immediate issue. I have done some (very) quick testing against a clone of a govCMS site already using the core Poll module and it appears as though it does not record anonymous votes after the first attempt. There appear to be similar experiences posted in the Drupal.org [issue queue](https://www.drupal.org/project/issues/pollanon?categories=All).

I'll try to find time to do some more thorough testing today.

aleayr:
Doesn't record anonymous votes after the first one, as in:
1. Multiple anonymous users can vote, but it restricts each of them to a single vote; or
2. After the first anonymous user votes, no more anonymous users can vote?

Cheers.

stooit:
Maybe both, after the initial testing from a single IP it recorded the first vote. After a cookie clear it did not record subsequent votes (although the UI to place a vote was correctly displayed rather than results).

As I said, was a very quick test. Will report back when I find time to test more thoroughly. In the meantime feel free to test against a vanilla govCMS install, would be useful to test a clean install vs. existing as I'm doing.

paulkilla:
This works. I've tested it locally. Deleting all cookies closing tab and re-opening allows to re-vote and the vote is recorded.
Edit: I wouldn't use AJAH.. AJAX method for recording votes though, even though it says a better user experience. It wasn't.

aleayr:
Yes, will do. Thanks.

stooit:
@paulkilla the issue I'm bumping into may be cache related -- can you confirm you are testing as an anonymous user with caching correctly configured?

paulkilla:
This is just locally, no caching. I'll leave it with you guys to test it with caching

fiasco:
Seems like this is our most viable option. So couple of things to check off:
- Module is no longer maintained. Perhaps we should consider taking some maintainership of it?
- This seems like specific functionality that should come with tests to ensure we don't regress this in the future?

aleayr:
Since the module is disabled by default, the tests needed to be written to enable this and test its functionality. As such, I think we'll progress sans tests at this immediate point, but we'll be looking further into robust tests in an upcoming distro sprint.

I'm also recommending to not take maintainership on at this point. If security updates are required, we can patch ourselves, but our focus as a small team should be on govCMS, and not a myriad of distractions.

@paulkilla Happy with the above? Are we good to progress to putting pollanon into testing?

paulkilla:
I'm happy with that. 

On 14 Oct 2015 10:37 am, Adam Leayr notifications@github.com wrote:

Since the module is disabled by default, the tests needed to be written to enable this and test its functionality. As such, I think we'll progress sans tests at this immediate point, but we'll be looking further into robust tests in an upcoming distro sprint.

I'm also recommending to not take maintainership on at this point. If security updates are required, we can patch ourselves, but our focus as a small team should be on govCMS, and not a myriad of distractions.

@paulkilla Happy with the above? Are we good to progress to putting pollanon into testing?

Issue #60, opened 2015-10-06
Drupal robots.txt and search engine indexing issue - request module to be added


Wongad:
The default robots.txt file in does not block or restrict search engine from Google or Bing from displaying your admin or user login pages.
e.g. Search on govCMS in [Google](https://www.google.com.au/?gws_rd=ssl#q=govcms+password) does not hide the /user, /user/password... etc from the [govCMS robots.txt](https://www.govcms.gov.au/robots.txt)

This is affecting our site at [PPSR](https://www.ppsr.gov.au) as we have user access to our [register application](https://transact.ppsr.gov.au/ppsr/Home?si=1) and search engine keeps coming up the admin/user page of Drupal, creating confusion.
This is an ongoing issue on Drupal api. https://api.drupal.org/api/drupal/robots.txt/7

As recommended by one of Acquia support, using the [robotstxt](https://www.drupal.org/project/robotstxt) module should be able to give us more control on the robots.txt: https://www.drupal.org/project/robotstxt

fiasco:
Being able to manage robots.txt per site seems like a good idea. We'll need to actively remove Drupal core's robot.txt during the govCMS build process which is doable also.
1. Add robotstxt to drupal-org.make file
2. Add as dependency to govcms install profile
3. Remove robots.txt from build in phing build file after drush make has completed running
4. Add test to assert robots.txt is not on the filesystem
5. Add test to asset that robots.txt is still accessible over HTTP.

stooit:
Happy to take care of this one, will submit a PR soon.

PR #67, opened 2015-10-09
Replace robots.txt with robotstxt module.

stooit:
Addresses #60.
1. Add robotstxt to drupal-org.make file
2. Add as dependency to govcms install profile
3. Remove robots.txt from docroot in phing build file after drush make has completed running
4. Add behat test to assert that robots.txt is still accessible over HTTP
5. Add behat test to assert that robots.txt is returned from robotstxt module and NOT robots.txt file on disk (checks for sitemap.xml only returned via robotstxt module)
6. Add update hook for existing govCMS sites to enable robotstxt module


fiasco:
thanks @srowlands!

Issue #45, opened 2015-09-01
installer UI alignment not pretty


kattekrab:
Just having a play with govCMS on simplytest me again, and this time remembered to log an issue.

I think the installer could do with a little theme love - the alignment is out of whack and doesn't give a good first impression.

Screenshot below.

![govcms-install-alignment](https://cloud.githubusercontent.com/assets/172755/9599158/d020c1d2-50d4-11e5-895e-694cf03232e5.png)

teamglenny:
Closing as has been made pretty(er) in #68 Will ultimatly be a work in progress

PR #68, opened 2015-10-09
Clean up display of installation forms.

stooit:
A minimal effort attempt as resolving #45.

Makes container wider and centers content within.

install-1
install-2
install-3
install-4


fiasco:
Definitely an improvement. Happy to commit as something we can iterate on and improve further as needed.

kattekrab:
Nice! That's much better. Thanks :)

PR #72, opened 2015-10-13
Upgrade search_api_solr to fix OR facets.

stooit:
Currently Facets are broken when using Search API solr when a facet is configured as an 'OR'.

This patch resolves this issue by upgrading to latest stable (1.9). [Discussion on Drupal.org issue](https://www.drupal.org/node/2290601)


fiasco:
The issue on d.o looks a mess. Its closed but with uncommitted patches? Does that mean this patch won't make it into Solr eventually?

stooit:
It looks like the patches have been committed in 1.9 stable. Testing 7.x-1.9 reveals no issues.

I'll change this PR to upgrade to 1.9 rather than patching 1.6

fiasco:
Looks like behat is taking to long to run for travis CI. That shouldn't stop this change from getting in.

PR #77, opened 2015-10-21
Add mailsystem and mimemail modules.

### LINKED ISSUE #66: Improve Webforms functionality with modules HTML Mail and Mime Mail
The current webform email functions of "Send e-mail as HTML" and "Include files as attachments" are not working as indicated.
webform

As recommended by Acquia support, I'm requesting the following two modules to be added to improve the webform email functionality.
- [HTML Mail](https://www.drupal.org/project/htmlmail)
- [Mime Mail](https://www.drupal.org/project/mimemail)

Currently we require attachments to be added in the email from the submission of the webform, in order for our CRM to store the necessary data. The two modules mentioned above will greatly enhance the webform email functions for further customisation.

stooit:
To resolve #66.

The mailsystem and mimemail modules are required to allow Webform to attach files to outgoing HTML-enabled emails. This is useful for submissions that need parsing and actioning by external systems (e.g CRM/IMS), and also for capturing content that may be potentially sensitive by bypassing storage in govCMS at all.

This is also a requirement for the Department of Communications in order to attach files to emails ingested by an internal system.


typhonius:
I think this will need rebasing against the latest HEAD to let it pass tests.

stooit:
Travis tests failing with `File govcms-7.x-2.x-dev.tar.gz_date=1445888940 is corrupt (wrong md5 checksum).`

pandaskii:
Hi guys,

Got same issue today

projects[govcms][version] = 2.x-dev (wrong md5 checksum occasionally from Drupal varnish)
projects[govcms][download][revision] = "c43f52e7b5792a54eb0bb16e10aaf29293fb7698" (give it a revision?)

stooit:
Thanks @jozhao, should be resolved by #85. Appreciate the input!

stooit:
test this

stooit:
To test this in the PR environment:
- http://govcms-pr77.52.64.233.163.xip.io/file-attachment-webform

Since we don't want to create a spam gateway outgoing mail will be captured by Mailhog. You can see access mailhog on port 8025:
- http://govcms-pr77.52.64.233.163.xip.io:8025/#

Checking the MIME tab will reveal downloadable attached files are working as expected.

fiasco:
- Should these modules be turned on by default?
- Do we need to have tests?

stooit:
- Should these modules be turned on by default?
  Since the option to 'attach as file attachments' is exposed via Webform but fundamentally broken without these modules I think the answer might be yes, actually.  Which means...
- Do we need to have tests?
  If we're providing this as an OOTB feature testing would make sense. I'm not sure _how_ we could test the receipt of an email with file attachments though.

fiasco:
Could https://mailinator.com/ work?

typhonius:
So if you look at .travis.yml we set the sendmail path to /bin/true in order to get through the install. 

We could perhaps implement a custom script as our sendmail path that gives a return value based on the input parameters and then test these modules based on that.

stooit:
@typhonius if we're moving toward behat tests running as a part of the PR environment we have a mail capture service (Mailhog -- http://govcms-pr77.52.64.233.163.xip.io:8025/) which we could also check.

The issue is these tests should work independently of whether it's Travis/Jenkins/local.. trying to keep something self contained so we can run the tests locally is probably important here.. Mailinator could be good for that reason but would need some further investigation

typhonius:
Definitely see the point in mailhog but as are we looking to install a system wide application (and maintain it) on Windows/Mac/Linux?

I'd prefer using something "proudly invented elsewhere" but would it be intrusive to install?

stooit:
Mailhog is only in place on PR environments being built to assist with testing functionality that requires sending mail (like this PR), I don't think it should be included in the govCMS codebase (seems like overkill to me).

This may just be one of those features that we can't accurately test with Behat alone, so maybe not worth overthinking it. Also enabling the 'send files as attachments' functionality would be enabled on a case-by-case basis, so it's also not a default we need to test for.

typhonius:
Although I did just RTFM and it says "MailHog runs without installation on multiple platforms." Maybe it _can_ check for attachments...

janeem:
Further clarification is sought on this request:
- Why are both the Mail system and Mime mail modules required?
- What is the business requirement?  That is, does the ability to upload attachments via Webform already exist?  Do these modules integrate with webform to extend the functionality so the files can be automatically emailed?
- The documentation for the Mail system module indicates there is a UI – I assume this is via the Drupal admin interface and available to certain roles?
- There is talk about using mailhog – is this a third party SMTP? It doesn’t appear to be a true SMTP so not clear on the value of introducing another third party in the mail process (recommend using a service hosted within AWS / greater assurance)
- Can the configuration of Mime mail restrict the path where files can be uploaded from

stooit:
> Why are both the Mail system and Mime mail modules required?

The MIME mail module is required to send HTML email, and include files as attachments. Mail system is a dependency of the mimemail module.

> What is the business requirement? That is, does the ability to upload attachments via Webform already exist? Do these modules integrate with webform to extend the functionality so the files can be automatically emailed?

Yes, the ability to upload attachments via Webform is already present in govCMS and currently in use. The option to attach files to outgoing email is also present, but does not function without the addition of these modules (see issue #66). These modules extend the existing functionality by making HTML email and file attachments function as expected.

The business requirement here is to prevent (potentially) sensitive file attachments from being stored within govCMS by having them emailed as attachments instead.

> The documentation for the Mail system module indicates there is a UI – I assume this is via the Drupal admin interface and available to certain roles?

There is an administrative Drupal UI to allow for configuring custom mail system interfaces, an advanced option that is unlikely to be used. By default it would not be accessible by anybody other than administrators.

> There is talk about using mailhog – is this a third party SMTP? It doesn’t appear to be a true SMTP so not clear on the value of introducing another third party in the mail process (recommend using a service hosted within AWS / greater assurance)

This was discussed purely as a way to test the functionality in the PR environment. There is no intent to add mailhog to the govCMS codebase.

> Can the configuration of Mime mail restrict the path where files can be uploaded from

This is beyond the scope of what these modules offer and the intent of this PR.

janeem:
Further discussion is required on the business requirement and the emailing of sensitive attachments in the clear via email. This will be raised with the business team.

janeem:
Thank you for the above response to my questions.  This is understood to extend the existing webforms functionality as used for public forms and feedback; and for attachments to be emailed as required (as outlined by the Dept of Comms requirement above).  Thanks for clarifying that mailhog is used for testing only and the admin UI is restricted to administrators.

fiasco:
Seems everyone is happy to merge this one?

barbun:
Any progress on this?

aleayr:
Final call before we merge this? Great work on this everyone.

PR #79, opened 2015-10-22
Implements a fix for https://bugs.php.net/bug.php?id=61286

typhonius:
Over in #67 we added robots.txt but due to https://bugs.php.net/bug.php?id=61286 the inbuilt PHP server has a special :panda_face: moment when we request a URI with a period in.

I've proposed a more appropriate fix upstream in Drush https://github.com/drush-ops/drush/pull/1699 although this will do for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,9 +49,9 @@ before_script:
   # Build the codebase
   - phing -f build/phing/build.xml build:local
 
-  # Run the Drush server
+  # Run the inbuilt PHP server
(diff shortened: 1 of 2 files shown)


typhonius:
Without variables provided by Drush, Drupal was like `wat` at the bare PHP -S implementation. I've created a router file which is essentially a bunch of stuff from Drush to stopgap this until Drush is fixed upstream :apple: :mortar_board:

PR #80, opened 2015-10-26
As a platform administrator I do not want agencies to be able to configure search api so that the search will not break

pandaskii:
Remove "search_api administer" permissions from all users except administrators.


fiasco:
@jozhao thanks for the contribution. Can you please specify why these permissions should be removed? Some context will help us understand why you think this is a good idea.

Thanks!

pandaskii:
@fiasco this is from the user story that "As a platform administrator I do not want agencies to be able to configure search api so that the search will not break". In this commit, we remove search_api admin permissions from all users except administrators by default.

fiasco:
Thanks @jozhao for the clarification. So this change is really around stability in the platform offering of govCMS. That being said, regardless, it makes sense to not give this capability to a site editor but I'm not sure about the site builder role. Wouldn't mind hearing the perspective of @srowlands  on that.

invisigoth:
@fiasco this issues was originally raised by Sean and he suggested "Remove permissions from all users except administrators." Reference: (GOVCMSDEV-103). Could you please review or communicate with Sean on this?

PR #82, opened 2015-10-27
exclude unpublished nodes from search index

pandaskii:
This commit enables the "Exclude unpublished nodes" in search API filters so that we can protect  unpublished draft content.

thanks


typhonius:
@jozhao are we able to get a test added to this PR as well to ensure we're protecting against regressions?

pandaskii:
screen shot 2015-10-27 at 2 22 10 pm

No problem mate. Trying to add them ASAP

Got some issues on local build process, govcms project is in my sites/all/modules folder

typhonius:
How're you building? Which phing commands are being used?

pandaskii:
@typhonius Got the reason and I will prepare the test codes tomorrow.

this is an interesting issue with Drush 7.10  if you don't specify a project type in your makefile, it will be automatically set to module, even if it's actually theme or profile https://github.com/drush-ops/drush/issues/1627 (have to run "composer update" instead of "composer install" for MAMP not having PHP extension pcntl - and this gave me Drush 7.10)

Excuse me for leave this here since I found many guys are using MAMP as govCMS playground. To fix it is quite easy, check this out https://github.com/drush-ops/drush/pull/1684/files 

On another hand, we may put this line into considerations. I will open an issue ticket on this if needed

```
projects[govcms][type] = profile
```

thanks mate

pandaskii:
Close this one, I will add a test and re-submit request.

Issue #84, opened 2015-10-27
Requests to updates.drupal.org return 404 Page not found which currently breaks build and test


pandaskii:
Downloading Drupal.org modules doesn't seem to work with Drush any more

There is an issue report in Drupal community queue now: https://www.drupal.org/node/2602264

typhonius:
Thanks for this. We experienced issues locally and presumed it was network issues.

typhonius:
Drush issue tracked https://github.com/drush-ops/drush/issues/1713

pandaskii:
@typhonius a big day for me, hah cannot handle with both Drush and Drupal suddenly change something at same day.

A interim way is to add curl --compressed in Drush

typhonius:
If nothing changes we can add that to DrushTask.php as a default option or to build.xml as a parameter passed through to Drush.

pandaskii:
Sounds good and nothing to lose. This is really a big change that many services depending on drush make, download will fail

typhonius:
This has been fixed on drupal.org by kicking an NFS mount :athletic_shoe: :computer:

PR #85, opened 2015-10-29
Resolve md5 checksum issue.

stooit:
Fixes the current build issue `File govcms-7.x-2.x-dev.tar.gz_date=1445888940 is corrupt (wrong md5 checksum).` by specifying a commit revision for the govCMS project.

Thanks @jozhao!

### DIFF
--- a/build-govcms.make
+++ b/build-govcms.make
@@ -6,3 +6,4 @@ includes[] = drupal-org-core.make
 
 ; Download the govCMS install profile and recursively build all its dependencies:
 projects[govcms][version] = 2.x-dev
+projects[govcms][download][revision] = "c43f52e7b5792a54eb0bb16e10aaf29293fb7698"


fiasco:
Is this a good idea? Doesn't this mean we have to essentially change this on every PR?

stooit:
This should only need changing each time a govCMS project release is made on drupal.org.

Regardless of if you are doing a make:local or make:remote Drush will attempt to download the govCMS project from d.o first (which is the point at which the build is currently failing). It then proceeds to overwrite the folder with the local build.

fiasco:
I've been doing that last few builds. The process I make is to first tag in github, then to tag in d.o and release. So you are saying that now I'll need to tag in d.o first. Then once the build is available, get the md5 checksum, add it to git, then tag in github to release?

Its not ideal. Is this all because of Travis? If we prioritise getting Jenkins up, will this go away?

stooit:
This is a new issue (we started running into it yesterday) and not Travis related.

If you download master and try to build (either local or remote) it fails with this error -- make sure your drush cache is cleared first.

It _may_ be related to the d.o infrastructure issue experienced on updates.d.o #84, but I don't see anything related in the d.o issue queue.

pandaskii:
Hi guys,

I submitted another option on this PR https://github.com/govCMS/govCMS/pull/86

Could be another option

Cheers

stooit:
#86 resolves

PR #88, opened 2015-10-29
Exclude unpublished nodes from search index

pandaskii:
This commit enables the "Exclude unpublished nodes" in search API filters so that we can protect unpublished draft content.

A behat search testing is added including two scenarios
1. Searching for a content that is published return result
2. Searching for a content that is NOT published return no results

@aleayr @invisigoth @ruwanl Once travis is happy, we are ready to merge it in next release.

### DIFF
--- a/build/tests/behat/features/search.feature
+++ b/build/tests/behat/features/search.feature
@@ -2,8 +2,63 @@ Feature: Search
 
   So that all users can search
   As an anonymous user I can use search
+  Exclude Unpublished content from search index
 
   @api @javascript
   Scenario: Search API block is present
     Given I am on the homepage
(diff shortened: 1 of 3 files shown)


fiasco:
This makes a lot of sense. Happy to commit it especially since its got tests! (love tests!) @jozhao can I suggest that you checkout [this behat video from DrupalCon](https://youtu.be/lfW2ngf5DgE?t=2174) which talks about writing good tests (presented by the guy who wrote behat). I don't want to let that hold this pull request back but its worth knowing.

pandaskii:
@fiasco happy to, appreciated for sharing the video :+1:

aleayr:
Thanks @jozhao and @fiasco. Good work :+1:

PR #90, opened 2015-10-29
As a platform administrator I do not want agencies to be able to configure search api so that the search will not break

pandaskii:
this issue was originally raised by Sean and he suggested "Remove permissions from all users except administrators." Reference: (GOVCMSDEV-103). 

@aleayr @invisigoth @ruwanl Once travis is happy, we are ready to merge it in next release.

### DIFF
--- a/modules/custom/govcms_permissions/govcms_permissions.perms.inc
+++ b/modules/custom/govcms_permissions/govcms_permissions.perms.inc
@@ -790,8 +790,6 @@ function govcms_permissions_base_permissions() {
       'administer search_api' => array(
         'roles' => array(
           0 => 'administrator',
-          1 => 'Site editor',
-          2 => 'Site builder',
         ),
       ),
     ),


typhonius:
Should this be tested to ensure a content editor cannot access the admin page?

pandaskii:
@typhonius good question, this is in my mind in past days :+1: 

Also when we add new module in govCMS, usually will bring new permissions and default configs

Should we have a permission testing feature in behat for all permissions defined in govcms_permissions.perms.inc? So in future, any permission change can follow this feature and add testing scenario

typhonius:
I don't think anyone would say no to more tests :smile:

pandaskii:
We should raise this in our team meeting :100: 

Not only permissions, but a testing plan including permissions, users, contents, search etc.

We can set up base testing features and encourage contributors to follow them

A structure could be:

user
-- rest password
-- sign in
-- etc

content
-- publish
-- moderate
-- etc

search
-- published
-- draft
-- etc

permissions
-- config search
-- delete content
-- etc

PR #91, opened 2015-10-29
Adding govcms favicon as default favicon to sit in the docroot

fiasco:
Browsers will always request /favicon.ico as a default location for favicons to load. Especially when HTML is not present to indicate otherwise (metatags). 

When we compile govCMS, we should place the govCMS favicon in place as a "catch all" and to prevent 404 requests of favicons can fall through to Drupal.

/favicons.ico is currently the most frequent 404 on the platform.

### DIFF
--- a/build/phing/build.xml
+++ b/build/phing/build.xml
@@ -217,6 +217,13 @@
         </if>
     </target>
 
+    <target name="mk-favicon"
+            description="Place the govCMS favicon in the docroot directory.">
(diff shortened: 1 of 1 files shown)


teamglenny:
Sounds good, @fiasco

fiasco:
We need to get a test environment for this. @govCMS-bot?

fiasco:
Looks good, favicon is present: http://govcms-pr91.52.64.233.163.xip.io/favicon.ico

PR #92, opened 2015-10-31
Fixes #3 by only checking required fields when not being installed.

### LINKED ISSUE #3: Database Configuration - required fields prevent form submission
Posted by tomz0r on July 14, 2015 at 4:13pm

/install.php?profile=govcms&locale=en

Having required="true" on the hidden form input fields is preventing the form from submitting. When you select MySQL as your database and fill in all the visible required fields, you also have to fill in dummy data in the PostgreSQL fields as well to be able to continue to the next step.

required_fields

typhonius:
Pull request to address needs of #3. This will need a manual review as our test procedure doesn't do a graphical install which is where this issue surfaces.

### DIFF
--- a/themes/govcms/govcms_zen/template.php
+++ b/themes/govcms/govcms_zen/template.php
@@ -201,7 +201,7 @@ function govcms_zen_form_required_marker($variables) {
  * Adds accessibility attributes.
  */
 function govcms_zen_preprocess_aria_invalid(&$variables) {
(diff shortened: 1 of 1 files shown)


pandaskii:
I have tested this in my local environment and works perfect

aleayr:
If you're happy with this one @jozhao, can you add the tech-vetted label to it, and we'll merge it in.

PR #95, opened 2015-11-11
More thorough testing of #59

stooit:
This is a dummy commit to spin up a PR environment in order to test the `pollanon` module introduced in PR #59.

Ignore for now, will report any findings back in the issue queue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # govCMS
 ***
 
+
 [![Build Status](https://api.travis-ci.org/govCMS/govCMS.svg?branch=master)](https://travis-ci.org/govCMS/govCMS)
 
 ## Installation


stooit:
### Testing process:
1. Enabled poll-anon
2. Created a new multi-choice poll (http://govcms-pr95.52.64.233.163.xip.io/example-poll-using-pollanon)
3. Enabling caching for anonymous users
4. Restrict polling to 5 votes per IP every 10m
5. Cleared cookies & browser cache and test for each of: Chrome 46.0.2490.86, Safari 9, Firefox 42, IE8, IE9, IE11, Edge
6. Attempt to submit poll without selecting option
7. Attempt to submit poll with option selected
8. Attempt to submit more than 5 votes from single IP within 10m period
9. Repeat testing on an ACSF environment behind varnish

### Notes:
1. Voting with caching enabled appears to work as expected when Varnish is not in the mix, however there are other issues noted below
2. Viewing results from a polls 'Results' tab displays cached data, whereas results provided after submission are correct as they are updated via js (possibly a non-issue)
3. _Does not work correctly behind Varnish._

### Current issues:
1. Submitting a poll without selecting an option causes a PDOException error as it does not correctly validate (dropped poll_view_voting_validate function)
2. IE9 & IE11 fail with "Cookie support required" error message
3. Results are cached unless you have voted and have a valid cookie
4. Broken behind varnish caching (similar issue to https://www.drupal.org/node/2552991):
5. Correctly accepts votes when it should, but results shown on tab and after voting are stale cached copy
6. Enabling 'AJAH' does not resolve (this appears non-functional).

### Proposed resolutions:
1. Test patch for PDOException (https://www.drupal.org/node/2483991)
2. Test patch for IE9 + IE11 (https://www.drupal.org/node/1580130)
3. Cached results for truly anon users (ie. those who have not voted) may be a non-issue
4. The varnish issue is a bigger problem as it makes the module practically worthless on an Acquia hosted environment (or any others that sit behind varnish).

fiasco:
Poll anon is now in govCMS so I think we can close this out.

stooit:
This testing environment was meant to show the issues with the poll_anon module; there are several issues that make this module unusable in its current state.

I'll port this over to the issues queue.

PR #97, opened 2015-11-17
Allow git to see compiled Drupal files

fiasco:
This is so that downstream projects don't get conflicts with govCMS's .gitignore ignoring their docroot directory.  This is an unnecessary blocker to upstream contribution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,11 +39,6 @@ Thumbs.db
 #######################
 .idea
 
-# Ignore Drupal stuff #
-########################
-docroot
-drushrc.php
-
 # Continuous Integration #
 #########################
 bin


.gitignore
```diff
@@ -39,11 +39,6 @@ Thumbs.db
 #######################
 .idea
 
-# Ignore Drupal stuff #
-########################
-docroot
-drushrc.php
```
typhonius, line 45: Are we removing this on purpose?

fiasco, line 45: Yup. The idea is, if I'm a downstream project using govCMS from source (because I'm also a contributor and its easier to contribute this way) this prevents me from tracking my docroot, where my compiled source is.

This feature is there only to hide these dirs from git but for downstream projects they will want to track them.

typhonius, line 45: But the drushrc.php line?

fiasco, line 45: Sure, that would need to be tracked in downstream projects also. Besides, govCMS doesn't compile drushrc.php for it to ignore right?

typhonius, line 45: govCMS doesn't use drushrc.php no but do you think users will need to commit that to their repo or should it be treated like any other local dev file and gitignored?

fiasco, line 45: I don't think drushrc.php is a dev file. Its an operational file that can provide context to drush commands that, for example, are used in cron runs.

typhonius, line 45: After reading the [documentation for drushrc.php](http://www.drupalcontrib.org/api/drupal/contributions!drush!examples!example.drushrc.php/7) I've changed my mind to agree with you :open_hands:

fiasco:
If you're happy with that, testbot is happy, are you happy to merge?

Issue #98, opened 2015-11-17
Remove ACSF build code from community distribution


fiasco:
Tools to build govCMS on ACSF don't need to be available in the open source repository as it doesn't really help any one. We should remove them and manage them in the correct version control systems.

PR #102, opened 2015-11-17
Support downstream building

fiasco:
This is a pull request for #101 to allow downstream projects to make customisations within a sites directory in the root of the repository. govCMS doesn't not track this sites folder so it will need additional documenting.

PR #103, opened 2015-11-18
Moved the location of the autodeploy script to a more appropriate place.

typhonius:
Starts to address #98


typhonius:
@fiasco the autodeploy script was something I am familiar with. For the rest of the files in the `build/acsf` directory here's what I'm thinking:
- `govcms_tweaks`: We should be handling distribution updates in `govcms.profile`, should this module be added in after govCMS is built but before it's deployed to ACSF
- `misc/favicon.ico`: Probably not a terrible idea to keep this in some way shape or form considering the number of 404s a docroot missing one can cause. Probably dealt with by #91 though.
- `misc/settings`: Similar story to `govcms_tweaks`?
- `patches`: These will need to form a part of the govCMS platform due to things like security considerations from forced HTTPS. They however are not required for local dev.
- `phing/acsf_post_make.build.xml`: Can be removed when `govcms_tweaks`, `patches` and `misc/settings` are removed.
- The other phing includes should be evaluated to determine the effect of their removal on the publicly built repo.
- As a follow up, the removed phing files can be removed from being called in the `build.xml`

fiasco:
#91 does fix the favicon issue and `build/acsf/misc/favicon.ico` is of the Australian crest which probably isn't appropriate for use outside the Australian government.

#102 will solve part of `phing/acsf_post_make_build.xml`. The rest can be ported to the ACSF repo itself. In fact, if you can take care of this autodeploy business, I can port the rest.

PR #118, opened 2015-12-02
Fixes a notice thrown on install in more recent PHP versions

typhonius:


fiasco:
This is a small coding bug so no product governance needed here. Merging.

PR #137, opened 2016-01-06
DRUPAL-SA-CONTRIB-2016-001: Update Field Group to v1.5

fiasco:
Field Group module enables you to group fields on entity forms and entity displays.

When adding a HTML element as group, the user has the option to add custom HTML attributes on the group. Via this option, a malicious user can embed scripts within the page, resulting in a Cross-site Scripting (XSS) vulnerability.

This vulnerability is mitigated by the fact that an attacker has to be able to configure field display settings, which usually needs a higher level permission such as Administer vocabularies and terms.


fiasco:
builds fine locally - this just looks like an issue with drush caching in travis.

PR #164, opened 2016-02-11
Update CKEditor to 4.5.7

mstrelan:
The current packaged version of CKEditor (4.4.7) is over a year old now and has had 9 releases since then. Version 4.4.8 includes a security update and there are plenty of enhancements and bugfixes in the 4.5.x branch, which is now used in Drupal 8 core.

One particular bug is fixed in 4.5.6 regarding files embedded from the media_wysiwyg plugin. It's probably an edge case and not worth describing, unless I need to further justify upgrading CKEditor.

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -179,7 +179,7 @@ libraries[html5placeholder][directory_name] = "html5placeholder"
 libraries[html5placeholder][type] = "library"
 
 libraries[ckeditor][download][type] = "get"
-libraries[ckeditor][download][url] = "http://download.cksource.com/CKEditor/CKEditor/CKEditor%204.4.7/ckeditor_4.4.7_full.zip"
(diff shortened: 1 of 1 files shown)


aleayr:
@jozhao Your thoughts/feedback?

pandaskii:
@aleayr This is a good idea with CKEditor 4.5.x

Drupal Community Documentation [Supported Editors Matrix](https://www.drupal.org/node/596966) also suggests Wysiwyg 7.x-2.x-dev can work with 4.5.7 full CKEditor package

It is better that we can also have a manual review

aleayr:
So @jozhao, are you saying we just need to do some more testing with this one, or saying it should be smooth since the Wysiwyg module supports 4.5.7?

pandaskii:
@aleayr it should be smooth and passed my local test. 

However, we need more testing from a tester to see if 
1. Compatible with the Media module/linkit etc
2. existing content is good with new version (such as node edit/view)

fiasco:

PR #166, opened 2016-02-16
Update metatag to 7.x-1.13

mstrelan:
The version of metatag in GovCMS is way out of date. We're having issues with caching of metatags which has been resolved in more recent versions.
- 7.x-1.5 adds an option to disable metatag output caching
- 7.x-1.6 resolves caching issues with Workbench Moderation
- 7.x-1.9 allows for better cache clearing logic


fiasco:
I agree, we need to get this in. But why did the tests fail?

mstrelan:
No idea! My last two seemed to fail randomly too.

fiasco:
The PR doesn't build for me locally for me. @aleayr the key thing we need to check here is if this upgrade will break any existing sites. So you're guys have any thoughts about how to determine that?

stooit:
`* [@self.4] Unable to patch metatag with support-all-attributes-1970362.pass_.patch.`

Will dig into why this might build locally vs. on the testing environments now.

fiasco:
Looks like this patch has not been rerolled in D7 https://www.drupal.org/node/1970362

aleayr:
To clarify, what's the work that needs doing here? I get that we're behind and should update, but is the latest module version failing both ours and Drupal's automated test suites?

Asides from having an approach to not break sites, @fiasco are we waiting for rerolls/tests to pass, or what's the next step?

fiasco:
Apparently we're tracking a patch for the metatag module which doesn't apply to the latest version:

```
 * [@self.4] Unable to patch metatag with                                [error]
```

We're tracking this patch here:

```
drupal-org.make:79:projects[metatag][patch][] = "https://www.drupal.org/files/issues/support-all-attributes-1970362.pass_.patch"
```

Which belongs to this issue: https://www.drupal.org/node/1970362
Which doesn't seem to have a reroll for the latest version.

aleayr:
@fiasco In relation to testing site issues, I think our intention was to roll up a release, tell customers what had been updated, update each test site and get site owners to test specifically for problems before going from there.

Feedback?

fiasco:
The metatag module provides non-visual functionality which can make it challenging to test. Especially for customers who do not have the permanent technical resources to conduct the testing.

Instead, if we had behat testing that tested the functionality from the metatag module we supported in govCMS, then we could update the module and be able to quickly verify that the supported functionality continued to work and we could release the update with confidence.

Aside from the testing issue, this PR needs to be rerolled and a decision needs to be made able the patch that no longer applies.

aleayr:
@mstrelan Are you able to reroll this to get it ready for merge? I'd like to merge it in since we're a fair way behind.

Noting behat testing for things like this (so we can release module updates faster), we should do a community and govCMS team blitz shortly on test cases.

aleayr:
Might be an idea to target to 1.14 as well since that is released now too.

fiasco:

Feng-Shui:
@fiasco was this closed because we didn't get the pull request re-submitted or for other reasons?

fiasco:
> This pull request needs to be re-submitted against the 7.x-2.x branch as the master branch has been removed per #226

It got closed due to the branch renaming - it was automatic.

PR #186, opened 2016-02-26
Add Email Field module 7.x-1.3 to govCMS distribution

barbun:
I think it is a good idea to add [Email Field](https://www.drupal.org/project/email) module to govCMS distribution as it provides valuable features such as validation, mailto links, encryption etc. It is also very well integrated and seems to respect accessibility.

In addition, email field type is a part of Drupal 8 core so it is just another tick of the box for this module to be included into GovCMS.


fiasco:
Thanks for the contribution @barbun!
I like the idea too but I think there maybe some security concerns. @aleayr can confirm, but I believe govCMS may only be allowed to host unclassified data and email addresses may be classified above that.

barbun:
No worries @fiasco, I figured there can be some concerns about that. On the other hand, you can still use a standard textfield for that purpose, the only difference would be some extra features that Email Field brings in (validation, display widgets, spam tools).

Perhaps, there could be a patch that would encrypt that field before storing it to the DB that would solve the issue?

fiasco:
Adding the email module would be govCMS endorsing the use of govCMS to store sensitive information. Using a textfield to do the same is beyond the control of what the govCMS project can control.

I'm not sure if encrypting data makes its unclassified. Again, I'd have to defer to @aleayr or @invisigoth to clarify.

Feng-Shui:
I've seen Departments use this field before, use case was structured contact information for various state government departments and NGOs.  It was all publicly available info and was collated into a single place for stakeholders.

Email field was used to ensure that email address were valid and to provide specific styling for the field.

In the above use case, there was no issue with classification.  Take the point that govCMS doesn't want to advocate for storing sensitive info, as with most things, depends on how you use it I guess :)

aleayr:
Sorry for missing this discussion, my filter I was using half-broke, so I've been missing some of these discussions.

Thanks a tonne for the contribution @barbun!

It comes down to a rather philosophical discussion about whether an email address constitutes personal information, and I'm currently seeking more information because we'd like to add things like email subscriptions and the like into the distribution, and this would be a natural compliment.

Currently though, while you can use just a text field and collect that information yourself, we're steered clear of specifically hinting/pointing people towards storing email addresses.

Hopefully I'll have some more news on this front not too far from now, but in the meantime, I might close this pull request for the meantime if no one has a specific objection to me doing so. Would definitely like to revisit this at some point though, because I think it's a useful addition if we can resolve the grey line of what information is personal and what isn't :)

fiasco:
Personally think we should keep the PR open until there is a real conclusion reached. I suspect this will come up again.

fiasco:

PR #187, opened 2016-02-29
Removed ACSF config from repo

fiasco:
I've removed references to ACSF as this stuff doesn't need to be tracked in the open project as it has zero open source value and just makes the build scripts more confusing then they need to be.


aleayr:
Do you have much to do with the build process on the Acquia side to reinstate these during build?

fiasco:
Already done as apart of the last build process. So its just cruft in this project now as the SaaS build doesn't require them any longer.

PR #190, opened 2016-03-03
Patched Context UI module to reduce memory usage

fiasco:
We've seen reports where Context UI is causing memory exhaustion issues. I've investigated the specific scenarios for existing govCMS customers and found that to be because Context UI loads in entire menus which can result in major memory consumption, even when those menu conditions are used.

This is an existing issue with Context and a patch for this specific issue exists from 4 years ago: https://www.drupal.org/node/873936. I've rerolled the patch so it applies to the latest version of Context which we run.


fiasco:
test this

fiasco:
#191 is required to fix this issue.

fiasco:
test this

fiasco:
I've added a second patch which allow you to use taxonomy loading over ajax instead of having to load it all in the admin page request. But you do need to set a variable for it to work:

```
drush vset context_condition_node_taxonomy_autocomplete 1
```

With the two patches in place and `context_condition_node_taxonomy_autocomplete` set to 1, the govCMS customer use case that was exceeding memory limits is now using 73.5M.

@gollyg, @paulkilla and @aleayr will be interested in this.

Feng-Shui:
`drush @fiasco friday-night --stop-working`

gollyg:
This issue will affect sites with reasonably sized menus and taxonomies. It is reasonable to expect that an increasing number of sites will encounter this over time.

I can confirm that with the two patches in place the memory usage decreases to a manageable level. @fiasco should the patch add an update hook to set that variable for existing sites?

Chris820:
I remember we had this problem with one of our sites a year or so ago.
A site with a large large menu structure would cause ContextUI to OOM, essentially making our Contexts uneditable.

So along with the patch that @fiasco mentioned, we also had to add this into settings.php
`$conf['menu_override_parent_selector'] = true;`

At least allowed us to edit the existing Contexts, but also meant that no new Menu based Contexts could then be created:
selection_745

Our longer term solution was to try to avoid making sites with single massive menu structures. And/or don't use Context

fiasco:
Thanks for the feedback @Chris820. We can selectively set `$conf['menu_override_parent_selector'] = true;` for sites affected on the platform. I think we've been lucky enough that sites don't use this context yet.

Agree with the guidance not to use Context. I'm not a fan either.

aleayr:
@Venkata-modgu Can we get this tested?

aleayr:
@jozhao Cant you do a code review on this? Thanks.

pandaskii:
Drupal 6/7 core has a built-in assumption that menus and vocabularies are small, so as context and tokens. For most projects it's a valid assumption, but the results can be disastrous while the projects have large amount of menus and terms. 

**'menu_override_parent_selector' variable**
In Drupal 7 core developer's mind, 'menu_override_parent_selector' variable is entirely superfluous. This variable is designed to allow contrib modules to provide more scalable pattern choosers. If this variable is set to TRUE, hook_form_alter is expected or other methods.

On another word, better to not change it if not having a method to provide pattern choosers. For example, a project with book module enabled. The book outline will be disarmed while this variable is set to TRUE.

There is an easy and safe way to achieve same result. Use hook_context_registry_alter()*

```
/**  
 * Implements hook_context_registry_alter(). 
 *
 * Examples ONLY NOT for PROD.
 */
 function govcms_core_context_registry_alter(&$registry)  { 
  // No Menu based Contexts.
  if (isset($registry['conditions']['menu'])) { 
    unset($registry['conditions']['menu']);   
  }  
  // No Node taxonomy based Contexts.
  if (isset($registry['conditions']['node_taxonomy'])) { 
    unset($registry['conditions']['node_taxonomy']);   
  }  
 // No Taxonomy term based Contexts.
 if (isset($registry['conditions']['taxonomy_term'])) { 
    unset($registry['conditions']['taxonomy_term']);   
  }   
  if (isset($registry['conditions']['defaultcontent'])) {    
    unset($registry['conditions']['defaultcontent']);   
  }  
 if (isset($registry['conditions']['book'])) {     
   unset($registry['conditions']['book']);   
  }   
  if (isset($registry['conditions']['bookroot'])) { 
    unset($registry['conditions']['bookroot']); 
  }
}
```

fiasco:
Thanks for the great suggestion @jozhao - I've implemented a slightly modified version.

stooit:
test this

pandaskii:
Only one issue now - it seems that all menus are flattened to 2 levels after applying patch. Please see the attached screenshots

snip20160411_1

snip20160411_2

Chris820:
I just applied @fiasco 's .patch to one of our sites. 
We've got some pretty big menus so figured it'll be a good real-world test.

The .patch alone gave a huge improvement over the ContextUI page load time. previously taking two minutes to load, down to a couple of seconds. Much better. Nice work on that!
(Although am also now getting the above mentioned trouble where the menu items are flattening down to one level).

Along with applying the .patch, I wrote the hook_context_registry_alter() trick into a new module (we're running the PaaS offering of GovCMS). The presence of this hook appears to prevent the creation of new contexts using one of the now unset conditions (which is fair enough). However, more importantly,  seems to break the reactions of existing Contexts that may already be using that condition.
It'll be worthwhile to point this out to users if planning to toggle any of the govcms_context_condition_$module_disabled variables.

Issue #106, opened 2015-11-25
As an Administrator, I want to give the Site Builder role permission to manage modules, but have a blacklist of modules they can't disable


aleayr:
We would like to allow Site Builders to enable/disable modules as required, however, some modules, such as Password Policy, should not be turned off.

Is there a way to implement a more granular approach to module management for Site Builders?

aleayr:
Another ticket I've since discovered, lists module_whitelist as a potential solution to this problem.

That ticket was for Site Editors, this for Site Builders.
Thoughts?

stooit:
`module_whitelist` is definitely a good solution here -- each time a new module is added to the distribution you'll want a process to evaluate and update the default whitelist rules.

invisigoth:
module_whitelist is certainly a good solution. However this module by itself does not allow users to then grand permissions of whitelisted modules to other roles/users. For example, if a site builder is allowed to turn on and off Views UI module, they should also be allowed to control which other users should be allowed to access Views UI module. This probably require another custom module to work with module_whitelist to provide all the features we wanted.

aleayr:
@invisigoth should have a prototype of this shortly.

invisigoth:
A new solution has been developed: https://www.drupal.org/project/module_permissions and is now being tested by our tester (GOVCMS-576). We will create a pull request when it passes the tests.

The module permissions module allows site administrator to define a subset of modules and allow users of another role to manage those selected modules. Those users can turn on and off those modules as well as manage permissions for them. It also allows delegates to enable dependencies.

Future improvements will be to allow delegates to configure modules in the managed list.

fiasco:
Feature is in but not enabled at present. Please contact Dept of Finance if you're interested in getting early access to this feature.

PR #194, opened 2016-03-09
Added GovCMS account security module

invisigoth:
(1) Addressed govCMS127 - As a user attempting to login, when I get my password incorrect a number of times, my status becomes blocked, and Drupal does not then send password reset emails.
(2) Addressed govCMS193 - Flood control status will be removed after a password reset link is followed
(3) Addressed issue - user name enumeration still possible on user profile page

Also a more generic message will be displayed when failed login limit is reached. This is consistent with current security practices for Government online services to avoid disclosure of the password management policy which could aid an attacker in developing more targeted attacks, e.g. denial of service, compromise of credentials, etc.

### DIFF
--- a/modules/custom/govcms_account_security/govcms_account_security.admin.inc
+++ b/modules/custom/govcms_account_security/govcms_account_security.admin.inc
@@ -53,9 +53,9 @@ function govcms_account_security_admin_settings() {
     '#type' => 'select',
     '#title' => t('HTTP code return to anonymous visits on user profile pages:'),
     '#options' => array(
-      '0' => t(variable_get('user_profile_page_code_anonymous_visit')),
+      '0' => t(variable_get('govcms_account_security_user_profile_page_code_anonymous_visit')),
     ),
-    '#default_value' => variable_get('user_profile_page_code_anonymous_visit'),
+    '#default_value' => variable_get('govcms_account_security_user_profile_page_code_anonymous_visit'),
     '#disabled' => TRUE,
   );
 
--- a/modules/custom/govcms_account_security/govcms_account_security.admin.inc
+++ b/modules/custom/govcms_account_security/govcms_account_security.admin.inc
@@ -2,9 +2,8 @@
 
 /**
  * @file
- * Contains admin paths for govCMS Account
+ * Contains admin paths for govCMS Account Security
  *
- * The original foundation for the govCMS distribution is aGov; the Drupal distribution created by PreviousNext to provide a core set of elements, functionality and features that can be used to develop government websites
  *
  * @copyright Copyright(c) 2015 Commonwealth of Australia as represented by Department of Finance
  * @license GPL v2 http://www.fsf.org/licensing/licenses/gpl.html
@@ -12,7 +11,7 @@
  */
 
 /**
- * Settings form for govcms password policy.
+ * Settings form for govcms account security.
  */
 function govcms_account_security_admin_settings() {
 
--- a/modules/custom/govcms_account_security/govcms_account_security.admin.inc
+++ b/modules/custom/govcms_account_security/govcms_account_security.admin.inc
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * @file
+ * Contains admin paths for govCMS Account
+ *
+ * The original foundation for the govCMS distribution is aGov; the Drupal distribution created by PreviousNext to provide a core set of elements, functionality and features that can be used to develop government websites
+ *
+ * @copyright Copyright(c) 2015 Commonwealth of Australia as represented by Department of Finance
+ * @license GPL v2 http://www.fsf.org/licensing/licenses/gpl.html
+ * @author Department of Finance
+ */
+
+/**
+ * Settings form for govcms password policy.
+ */
+function govcms_account_security_admin_settings() {
+
+  $form = array();
+
+  drupal_set_message(t("Settings are for review only and cannot be changed."), 'warning');
+
+  $form['govcms_account_security_user_failed_login_user_limit'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Number of failed login attemps allowed before blocking an account:'),
+    '#options' => array(
+      '0' => t(variable_get('user_failed_login_user_limit')),
+    ),
+    '#default_value' => variable_get('user_failed_login_user_limit'),
+    '#disabled' => TRUE,
+  );
+
+  $form['govcms_account_security_user_failed_login_user_window'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Block an account for after failed login limit is reached:'),
+    '#options' => array(
+      '0' => t(variable_get('user_failed_login_user_window').' seconds'),
+    ),
+    '#default_value' => variable_get('user_failed_login_user_window').' seconds',
+    '#disabled' => TRUE,
+  );
+
+  $form['govcms_account_security_user_failed_login_identifier_uid_only'] = array(
+    '#type' => 'select',
+    '#title' => t('Flood control event identifer:'),
+    '#options' => array(
+      '0' => t(variable_get('user_failed_login_identifier_uid_only')?'User ID only':'Both user ID and IP address'),
+    ),
+    '#default_value' => variable_get('user_failed_login_identifier_uid_only')?'User ID only':'Both user ID and IP address',
+    '#disabled' => TRUE,
+  );
+
+  $form['govcms_account_security_user_profile_page_code_anonymous_visit'] = array(
+    '#type' => 'select',
+    '#title' => t('HTTP code return to anonymous visits on user profile pages:'),
+    '#options' => array(
+      '0' => t(variable_get('user_profile_page_code_anonymous_visit')),
+    ),
+    '#default_value' => variable_get('user_profile_page_code_anonymous_visit'),
+    '#disabled' => TRUE,
+  );
+
+  return $form;
+}
--- a/modules/custom/govcms_account_security/govcms_account_security.install
+++ b/modules/custom/govcms_account_security/govcms_account_security.install
@@ -1,0 +1,72 @@
+<?php
+/**
+ * @file
+ * Install file for account lockout.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function govcms_account_security_install() {
+  // ISM 2015, control: 1403, accounts are locked after a maximum of 5 failed logon attempts
+  variable_set('user_failed_login_user_limit', 5);
+  // Lock the offending account for 315400000 seconds (10 years)
+  variable_set('user_failed_login_user_window', 315400000);
+  // Register flood events based on the uid only, see user.module
+  variable_set('user_failed_login_identifier_uid_only', TRUE);
+  // Set HTTP status code returned to anonymous visits on user profile pages
+  variable_set('user_profile_page_code_anonymous_visit', '403');
+  // An accounts with expired password will be forced to change their password upon the next login
+  variable_set('password_policy_block', 1);
+
+}
+
+/**
+ * Implements hook_enable().
+ */
+function govcms_account_security_enable() {
+  $modules_to_disable = array();
+  if(module_exists('login_security')) {
+    $modules_to_disable[] = 'login_security';
+  }
+  if(module_exists('flood_control')) {
+    $modules_to_disable[] = 'flood_control';
+  }
+  if (!empty($modules_to_disable)) {
+    module_disable($modules_to_disable);
+    drupal_set_message(t("Login Security and Flood Control modules are automatically disabled."), 'warning');
+  }
+}
+
+/**
+ * Implements hook_requirements().
+ */
+function govcms_account_security_requirements($phase) {
+  $requirements = array();
+  $t = get_t();
+  if ($phase == 'runtime' && (module_exists('login_security') || module_exists('flood_control'))) {
+    $requirements['govcms_account_security'] = array(
+      'title' => $t('Login Security and Flood Contol'),
+      'value' => $t('Enabled'),
+      'description' => $t('Login Security and Flood Control modules need to be disabled.'),
+      'severity' => REQUIREMENT_INFO,
+    );
+  }
+  return $requirements;
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function govcms_account_security_uninstall() {
+
+  // Delete the variables to use the value set by Drupal's user module
+
+  // Drupal's default value is 5
+  variable_del('user_failed_login_user_limit');
+  // Drupal's default value is 21600
+  variable_del('user_failed_login_user_window');
+  // Delete account security specific var
+  variable_del('user_profile_page_code_anonymous_visit');
+
+}
--- a/modules/custom/govcms_account_security/govcms_account_security.install
+++ b/modules/custom/govcms_account_security/govcms_account_security.install
@@ -1,7 +1,7 @@
 <?php
 /**
  * @file
- * Install file for account lockout.
+ * Install file for govcms account security.
  */
 
 /**
--- a/modules/custom/govcms_account_security/govcms_account_security.install
+++ b/modules/custom/govcms_account_security/govcms_account_security.install
@@ -15,7 +15,7 @@ function govcms_account_security_install() {
   // Register flood events based on the uid only, see user.module
   variable_set('user_failed_login_identifier_uid_only', TRUE);
   // Set HTTP status code returned to anonymous visits on user profile pages
-  variable_set('user_profile_page_code_anonymous_visit', '403');
+  variable_set('govcms_account_security_user_profile_page_code_anonymous_visit', '403');
   // An accounts with expired password will be forced to change their password upon the next login
   variable_set('password_policy_block', 1);
 
@@ -67,6 +67,6 @@ function govcms_account_security_uninstall() {
   // Drupal's default value is 21600
   variable_del('user_failed_login_user_window');
   // Delete account security specific var
-  variable_del('user_profile_page_code_anonymous_visit');
+  variable_del('govcms_account_security_user_profile_page_code_anonymous_visit');
 
 }
--- a/modules/custom/govcms_account_security/govcms_account_security.module
+++ b/modules/custom/govcms_account_security/govcms_account_security.module
@@ -8,7 +8,7 @@
 /**
  * Implements hook_menu_alter().
  */
-function govcms_account_security_menu(&$items) {
+function govcms_account_security_menu() {
   // We want to create a review settings link visiable under admin/config/people
   $items['admin/config/people/account_security'] = array(
     'title' => 'Review account security settings',
@@ -19,7 +19,7 @@ function govcms_account_security_menu(&$items) {
     'file' => drupal_get_path('module', 'govcms_account_security') . '/govcms_account_security.admin.inc',
     'description' => 'Review govCMS account security settings.',
   );
-
+  return $items;
 }
 
 /**
--- a/modules/custom/govcms_account_security/govcms_account_security.module
+++ b/modules/custom/govcms_account_security/govcms_account_security.module
@@ -6,7 +6,7 @@
  */
 
 /**
- * Implements hook_menu_alter().
+ * Implements hook_menu().
  */
 function govcms_account_security_menu() {
   // We want to create a review settings link visiable under admin/config/people
@@ -16,7 +16,7 @@ function govcms_account_security_menu() {
     'page arguments' => array('govcms_account_security_admin_settings'),
     'access callback' => 'user_access',
     'access arguments' => array('administer users'),
-    'file' => drupal_get_path('module', 'govcms_account_security') . '/govcms_account_security.admin.inc',
+    'file' => 'govcms_account_security.admin.inc',
     'description' => 'Review govCMS account security settings.',
   );
   return $items;
--- a/modules/custom/govcms_account_security/govcms_account_security.module
+++ b/modules/custom/govcms_account_security/govcms_account_security.module
@@ -1,0 +1,112 @@
+<?php
+
+/**
+ * @file
+ * govCMS Account Security
+ */
+
+/**
+ * Implements hook_menu_alter().
+ */
+function govcms_account_security_menu_alter(&$items) {
+
+  $items['admin/config/people/account_security'] = array(
+    'title' => 'Review account security settings',
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('govcms_account_security_admin_settings'),
+    'access callback' => 'user_access',
+    'access arguments' => array('administer users'),
+    'file' => drupal_get_path('module', 'govcms_account_security') . '/govcms_account_security.admin.inc',
+    'description' => 'Review govCMS account security settings.',
+  );
+
+}
+
+/**
+ * Implementation of hook_form_FORM_ID_alter()
+ */
+function govcms_account_security_form_user_pass_reset_alter(&$form, &$form_state, $form_id) {
+  if(isset($form_state['build_info']['args'][0]) && $form_state['build_info']['args'][0] > 0) {
+    $uid = $form_state['build_info']['args'][0];
+    // Check if the user attempting to reset their password is flood controlled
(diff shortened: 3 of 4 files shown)


fiasco:
I see this has a dependancy on password policy module which we run v1.11 of but the recommended version is v2. Add this capability (which I'm for) will add technical debt to migrate to password policy v2 if we wanted to. Or will it be all that complex?

invisigoth:
@fiasco the dependancy "password policy" is declared because we set variable_set('password_policy_block', 1);
This is to address another security issue reported earlier this year.

I suggest we could discuss the password policy module upgrade path separately.

fiasco:
Agreed, sounds like it is separate enough to be a non-issue.

fiasco:
@invisigoth could you add some behat tests that verify that your module does what its suppose to do? Since this PR adds explicit behavioural features to govCMS, it is a prime candidate to be testable.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Implements hook_form_alter().
+ */
+function govcms_account_security_form_alter(&$form, &$form_state, $form_id) {
+  switch ($form_id) {
+    case 'user_login':
+    case 'user_login_block':
+      $user_login_final_validate_index = array_search('user_login_final_validate', $form['#validate']);
+      if ($user_login_final_validate_index >= 0) {
+        $form['#validate'][$user_login_final_validate_index] = 'govcms_account_security_final_validate';
```
seanhamlin, line 59: This seems like this should have extensive comments as to why we are unsetting Drupal core's default validation function. The had better be an awesome reason for this too. I would also be interested in knowing we can cannot do what is standard practice, being to append our validation function to the end of the array().

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+function govcms_account_security_final_validate($form, &$form_state) {
+  if (empty($form_state['uid'])) {
+    // Always register an IP-based failed login event.
+    flood_register_event('failed_login_attempt_ip', variable_get('user_failed_login_ip_window', 3600));
+    // Register a per-user failed login event.
+    if (isset($form_state['flood_control_user_identifier'])) {
+      flood_register_event('failed_login_attempt_user', variable_get('user_failed_login_user_window', 21600), $form_state['flood_control_user_identifier']);
+    }
+    if (isset($form_state['flood_control_triggered'])) {
+        form_set_error('name', t('Unrecognised username or password. <a href="@password">Have you forgotten your password?</a>', array('@password' => url('user/password', array('query' => array('name' => $form_state['values']['name']))))));
```
seanhamlin: spacing issues. Can you run PHPCS over this module?

modules/custom/govcms_account_security/govcms_account_security.admin.inc
```diff
@@ -0,0 +1,63 @@
...
+    ),
+    '#default_value' => variable_get('user_failed_login_identifier_uid_only')?'User ID only':'Both user ID and IP address',
+    '#disabled' => TRUE,
+  );
+
+  $form['govcms_account_security_user_profile_page_code_anonymous_visit'] = array(
+    '#type' => 'select',
+    '#title' => t('HTTP code return to anonymous visits on user profile pages:'),
+    '#options' => array(
+      '0' => t(variable_get('user_profile_page_code_anonymous_visit')),
```
seanhamlin: Should the variables be namespaced with the prefix "govcms_account_security"?

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+    // log in and out more than once in an hour.
+    flood_clear_event('failed_login_attempt_user', $form_state['flood_control_user_identifier']);
+  }
+}
+
+/**
+ * Implements hook_init().
+ */
+function govcms_account_security_init() {
+  if (!user_is_logged_in()) {
```
seanhamlin: there is a function "user_is_anonymous"

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+    // Clear past failures for this user so as not to block a user who might
+    // log in and out more than once in an hour.
+    flood_clear_event('failed_login_attempt_user', $form_state['flood_control_user_identifier']);
+  }
+}
+
+/**
+ * Implements hook_init().
+ */
+function govcms_account_security_init() {
```
seanhamlin, line 102: Seems like a lot of work to be doing in hook_init(). I don't know if this is the best approach. In any case, we should be excluding cli commands from this.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+  if (!user_is_logged_in()) {
+    $alias = drupal_get_path_alias(current_path());
+    if (drupal_match_path($alias, 'users/*')) {
+      // Throw an access denied error to any user profile page.
+      // By default we don't want to throw a 404 code because search 404 will then expose the user ID if the path actually exists
```
seanhamlin: Comment wrap.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @file
+ * govCMS Account Security
+ */
+
+/**
+ * Implements hook_menu_alter().
+ */
+function govcms_account_security_menu_alter(&$items) {
```
seanhamlin: Use hook_menu(), not hook_menu_alter().

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,112 @@
...
+ */
+function govcms_account_security_form_user_pass_reset_alter(&$form, &$form_state, $form_id) {
+  if(isset($form_state['build_info']['args'][0]) && $form_state['build_info']['args'][0] > 0) {
+    $uid = $form_state['build_info']['args'][0];
+    // Check if the user attempting to reset their password is flood controlled
+    if (!flood_is_allowed('failed_login_attempt_user', variable_get('user_failed_login_user_limit'), variable_get('user_failed_login_user_window'), $uid)) {
+      $timestamp = $form_state['build_info']['args'][1];
+      $hashed_pass = $form_state['build_info']['args'][2];
+      $users = user_load_multiple(array($uid), array('status' => '1'));
+      $current = REQUEST_TIME;
```
seanhamlin, line 36: REQUEST_TIME is the standard Drupal constant, no need to assign it to another (more obscure) variable.

seanhamlin:
Will this module stand independent of govCMS? i.e. will it be open sourced on d.o?

invisigoth:
@seanhamlin 

> This seems like this should have extensive comments as to why we are unsetting Drupal core's default validation function. The had better be an awesome reason for this too. 

This is specifically requested by the security team and the rationale is:

> In relation to the generic message this is consistent with current security practices for Government online services to avoid disclosure of the password management policy which could aid an attacker in developing more targeted attacks, e.g. denial of service, compromise of credentials, etc.

@aleayr, will you be able to provide an answer to:

> Will this module stand independent of govCMS? i.e. will it be open sourced on d.o?

fiasco:
Can you avoid disclosure of the password management policy in open source? This is the classic case of the saying "Security through obscurity is not security".

maxious:
> Control: 0421; Revision: 4; Updated: Apr-15; Applicability: UD, P, C, S; Compliance: must; Authority: AA
> Agencies using passphrases as the sole method of authentication **must** enforce the following
> passphrase policy...
> http://www.asd.gov.au/publications/Information_Security_Manual_2015_Controls.pdf

modules/custom/govcms_account_security/govcms_account_security.module
invisigoth: Please refer to user.module

modules/custom/govcms_account_security/govcms_account_security.module
invisigoth: Please refer to user.module

modules/custom/govcms_account_security/govcms_account_security.module
invisigoth, line 36: Please refer to user.pages.inc

modules/custom/govcms_account_security/govcms_account_security.module
pandaskii: The functions are essentially opposites.

The only real difference is the user_is_anonymous() function has an additional check for $GLOBALS['menu_admin'] to allow menu admins to be able to administer menu links that are visible to anonymous users only.

It's used by drupal_valid_path() as well as the menu_overview_form() function

seanhamlin: Right gotcha, I just thought this would be easier to read.

modules/custom/govcms_account_security/govcms_account_security.module
seanhamlin: Fairly sure Drupal core does not contain comments that exceed 80 chars. It is in the coding standards.

invisigoth: I see what you mean.

modules/custom/govcms_account_security/govcms_account_security.admin.inc
invisigoth: Good idea will do

modules/custom/govcms_account_security/govcms_account_security.module
invisigoth: Will fix

aleayr:
@invisigoth @seanhamlin 

> Will this module stand independent of govCMS? i.e. will it be open sourced on d.o?
> It's a pretty specific module, covering a very specific use case of config and other modules clashing. I'm not saying we shouldn't, but does it add any value when it's so specific to this govCMS use case?

@fiasco 

> Can you avoid disclosure of the password management policy in open source? This is the classic case of the saying "Security through obscurity is not security".

Definitely, and I want to make it clear we're not advocating that. We understand the install profile contains these settings, but our advice from penetration testers is that it is best practice not to give away more information than you have to about your security policies, and to keep error messages generic.

From a UX perspective, I totally understand telling the user they've been locked out, it does come down to a risk strategy about how much information is given away.

The option to reimplement that message (and the associated overwriting of core functionality), was not arrived at lightly, and I understand and respect the opinions expressed thus far, however, while we all agree it's untidy, what are the risks in your mind of overwriting this small part?

Bearing in mind, I do need to get this PR merged and into a release to solve an issue that stops people logging in when they've been flood controlled, so would like to get on the same page as soon as possible :)

pandaskii:
@Venkata-modgu Please test this pull request, thanks (cc @aleayr)

Venkata-modgu:
Test passed.

modules/custom/govcms_account_security/govcms_account_security.admin.inc
```diff
@@ -0,0 +1,63 @@
...
+    ),
+    '#default_value' => variable_get('user_failed_login_user_limit'),
+    '#disabled' => TRUE,
+  );
+
+  $form['govcms_account_security_user_failed_login_user_window'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Block an account for after failed login limit is reached:'),
+    '#options' => array(
+      '0' => t(variable_get('user_failed_login_user_window').' seconds'),
```
fiasco, line 36: This is a bad use of the `t` function as you can't translate text that is dynamic. Also variable_get doesn't have a default which it should.

This should be written like this:

```
t('@number seconds', array('@number' => variable_get('user_failed_login_user_window', 30));
```

modules/custom/govcms_account_security/govcms_account_security.module
seanhamlin, line 59: After reading https://api.drupal.org/api/drupal/modules%21user%21user.module/function/user_login_final_validate/7 in more detail, it appears that this function only renders error messages and registers flood events. 

After reading govcms_account_security_final_validate (below), I suspect the only changes are what strings are presented to the end user in the form of errors. 

I am keen to understand why we don't simply alter the strings using (core) localisation? An example on how to do this in the the core settings.php file:

e.g. 

```
$conf['locale_custom_strings_en'][''] = array(
  'Sorry, too many failed login attempts from your IP address. This IP address is temporarily blocked. Try again later or <a href="@url">request a new password</a>.' => 'Unrecognised username or password. <a href="@password">Have you forgotten your password?</a>',
);
```

Keen to hear what other people think about this.

modules/custom/govcms_account_security/govcms_account_security.install
```diff
@@ -0,0 +1,72 @@
...
+  $t = get_t();
+  if ($phase == 'runtime' && (module_exists('login_security') || module_exists('flood_control'))) {
+    $requirements['govcms_account_security'] = array(
+      'title' => $t('Login Security and Flood Contol'),
+      'value' => $t('Enabled'),
+      'description' => $t('Login Security and Flood Control modules need to be disabled.'),
+      'severity' => REQUIREMENT_INFO,
+    );
+  }
+  return $requirements;
```
seanhamlin, line 55: I am guessing there is nothing stopping someone changing the values e.g. 'user_failed_login_user_limit' to another value after installing this module (e.g. using drush). Would it make sense to:
- hard code the values in settings.php (this enforces the security policy)
- add a status report check to ensure the variables are set correctly

Keen to hear thoughts on this from everyone.

pandaskii, line 55: Settings file is a good thinking from hosting side or single site. From
distribution level it is out of scope as some govcms sites are hosted by
themselves or other hosting providers.

Status check for variable if someone change value using drush is not quite
necessary. IMO if this user can use drush, a warning cannot stop him :)

modules/custom/govcms_account_security/govcms_account_security.module
fiasco, line 102: I don't understand why this is being done in the first place? You can restrict anon users from accessing user pages without needing to write code.

I agree, this shouldn't be done in hook_init either. That gets called on every page and the criteria for this code to execute is going to occur 0.0001% of the time. This sort of logic would be better handled with hook_menu_alter:
- Implement hook_menu_alter
- Override access callback for user/% paths and set it to your own callback that does two things:
  1. Checks the user is logged in.
  2. Calls the original access callback, [user_view_access](https://api.drupal.org/api/drupal/modules%21user%21user.module/function/user_view_access/7)

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+/**
+ * Implements hook_init().
+ */
+function govcms_account_security_init() {
+  if (!user_is_logged_in()) {
+    $alias = drupal_get_path_alias(current_path());
+    if (drupal_match_path($alias, 'users/*')) {
```
fiasco: Why evaluate a condition based on a path alias instead of its source path? Couldn't govCMS users get around this security feature by just changing the default URL aliases patterns for user entities?

modules/custom/govcms_account_security/govcms_account_security.install
invisigoth, line 55: It is intentional to allow changing the failed login limit using drush. It is understood those with drush access should be trusted users and they can do a lot more than changing a failed login limit.

modules/custom/govcms_account_security/govcms_account_security.admin.inc
invisigoth, line 36: It is intentional to not to have a default value set.

modules/custom/govcms_account_security/govcms_account_security.module
invisigoth: @fiasco good suggestion. will look into it. the idea here was originally to only patch identified issues.

modules/custom/govcms_account_security/govcms_account_security.module
typhonius, line 102: I agree with @fiasco - why not just alter the permission to 'view user profiles' in the govCMS distribution?

modules/custom/govcms_account_security/govcms_account_security.install
```diff
@@ -0,0 +1,72 @@
...
+function govcms_account_security_enable() {
+  $modules_to_disable = array();
+  if(module_exists('login_security')) {
+    $modules_to_disable[] = 'login_security';
+  }
+  if(module_exists('flood_control')) {
+    $modules_to_disable[] = 'flood_control';
+  }
+  if (!empty($modules_to_disable)) {
+    module_disable($modules_to_disable);
```
typhonius, line 36: Is automatically disabling modules a site relies on good practice? Should it be the case that this module can't be enabled if those modules are enabled. This could be achieved by implementing a check in the `install` phase of `hook_requirements()`

fiasco, line 36: Thats a good point.We now have an [audit tool](https://github.com/seanhamlin/site-audit) where we can enforce this on the platoform. Advisory is probably the best approach for the distribution. #106 also looks like it will help prevent these modules from being enabled.

pandaskii, line 36: @typhonius @fiasco How about uninstall these two modules in next release after this module is enabled? Cut them from root :)

modules/custom/govcms_account_security/govcms_account_security.module
pandaskii: Here is issue background:

Currently:

http://194.govcms.xyz/user/100 - invalid user -> return 404

http://194.govcms.xyz/user/1 - valid user -> return 403 -> redirect to user login

We can use this to get the status of the user from user id and username. For example, 

http://194.govcms.xyz/users/admin - 403 valid user with a possible username admin

http://194.govcms.xyz/users/test - 404 invalid user, try another one from dict.

Therefore, we want a same 403 or 404 whatever status of user

@fiasco raised a good point here.

fiasco: What if we removed 403's from govCMS with this module? https://www.drupal.org/project/m4032404
This is a security practice that is also used by github to not disclose private repository information.

modules/custom/govcms_account_security/govcms_account_security.install
invisigoth, line 36: The "login_security" and "flood_control" modules are already disabled after the security issues identified and reported in November last year. 

It is our intention to not to enable this account security module automatically at this stage and only manually enable it on the sites where issues have been reported. We will then enable this module by default after a review.

modules/custom/govcms_account_security/govcms_account_security.module
aleayr, line 59: Yep, essentially conforming to our security teams recommendation for error messages on government systems. Also using it to register flood events as you noted. Currently, using this approach in order to provide translations rather than just English in the localisations config.

I'm happy a considered approach has been taken to this.

modules/custom/govcms_account_security/govcms_account_security.module
aleayr, line 102: It's because that's too late in the load process. Scanning user ids will return either a 404, or a 403, which gives out which users are in the system. We need to get into the load process early enough to be able to force all user/\* requests to be 403s.

I'm satisfied even though this is run each page load, that the overhead is very small (simple if checking the path), and the result will be cached in any case.

aleayr:
I'm satisfied with this approach, and there has been some really good and appreciated discussion on this.

Unless there are any further outstanding discussion items, I'm intending to merge it in for the next release.

modules/custom/govcms_account_security/govcms_account_security.admin.inc
fiasco, line 36: Is an `#options` key even valid here? I've not seen it used in the context of a textfield and didn't think textfield widgets had any support for it?

modules/custom/govcms_account_security/govcms_account_security.install
fiasco, line 36: So if modules on the SaaS platform have already been disabled and resolved. This code is purely for non-SaaS users like @Feng-Shui so I'd be interested in his thoughts as to whether, as distribution maintainers, we should make these decisions for them. 

In settings where govCMS maybe used with custom modules and additional contrib, they may have dependancies on modules we've historically had enabled. So without dependancy checking, this can be dangerous.

IMO, we should use hook_requirements to advise users to disable these modules but not enforce it.

modules/custom/govcms_account_security/govcms_account_security.module
fiasco, line 102: While this approach has a small overhead when looked at in isolation, using this architecture as a best practice will snowball the overhead after the hook_init strategy is used for several other scenarios.

> We need to get into the load process early enough to be able to force all user/\* requests to be 403s.

I don't think we're doing that with the current code. It looks to me like it only cares about /user/%uid and omits consideration to 403s from /user/%uid/edit and any other user tabs added by contrib modules.

> It's because that's too late in the load process.

A menu access callback is what controls whether a 403 is issued or not. So it allows us to prevent that from happening and issuing a 404 instead. Furthermore, it executes only in the context of the matched menu item.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -0,0 +1,127 @@
...
+      // automatically update existing user aliases when the pattern is
+      // changed, aliases generated using the old pattern will not be
+      // protected until a bulk update is performed.
+      if (strpos($user_alias_pattern, '[user:name]') === FALSE) {
+        $user_alias_pattern = 'users/*';
+      } else {
+        $user_alias_pattern = str_replace('[user:name]', '*', $user_alias_pattern);
+      }
+    }
+    if (drupal_match_path($alias, $user_alias_pattern)) {
```
fiasco, line 117: I think this code should be accompanied with tests. If this user pattern matches an alias that is not a user view page, it could essentially break the site which no really explanation as to why. We should have tests that ensure that can't be the case.

modules/custom/govcms_account_security/govcms_account_security.install
invisigoth, line 36: The one simple reason behind disabling the "login_security" module is that it does not fully comply with least one control (Compliance: must; Authority: AA) in ISM 2015. This means for a govt website to maintain its compliance with ISM, it should implement this control using an alternative method rather than using "login_security". 

As a distribution intended for government use, we have the utmost responsibility to maintain the compliance.

As far as I'm aware, permissions on the SaaS platform have not always been correctly given by some developers and at least one agency had revoked the changes to mitigate the risks. Now as the permissions on this site been corrected, we need to make sure the fixes are re-applied and enforced.

fiasco, line 36: It definately makes sense to enforce this for SaaS users. But since the distribution is used by more than just the SaaS platform, it makes sense to manage the enforcement though other means.

I don't think we can assume we have responsibility over the use of govCMS when it is used outside of our control. In those settings, agencies can augment the distribution in ways we can't and shouldn't cater for. But we shouldn't break that capability either.

Thats why I'm advocating for strong advisory in this scenario rather than forcing a potentially breaking change.

aleayr:
I'm satisfied that this has been tested, it isn't going to be on by default at this stage, and we will be encouraging customers to conduct testing on their own sites to compliment our own testing. Before it gets enabled by default in a future release, I will ensure appropriate documentation is written as to the implications of this change.

xtfer:
I'm not sure about govcms_account_security_init()... I can see multiple issues with it. 
- Looks like it clobbers all user paths for anonymous users regardless of permissions.
- Doesn't fix the enumeration problem. Non-existing accounts return a 404, so this should do the same thing when it blocks access. Additionally, there may be circumstances where you want to allow access to those paths.
- Preventing access in hook_init() is generally considered bad practice. Was overriding the `access callback` for the menu router considered?
- As it stands, this wont run on cached paths, or if pathauto is not in use.
- There are grammar issues.

Messing with user access, especially on /user/\* paths, is a bit of a minefield. This probably should have more testing before being readily available.

Issue #130, opened 2015-12-15
As a website administrator I want to export views so that I can use this data in other ways. 


WebProject2015:
Due to govcms being such a powerful tool for managing content, there is often a business requirement to use this managed content outside of the system itself. Views allow lists and tables to be generated programatically in ways which are not possible using internal tools, and websites often combined information from various disperate areas of an organisation.  As such there are often requests to web teams for reports, or for excel versions of dynamically generated tables from the site. This will be a universal challenge for all agency web teams.  

 Examples might be a request from management for a list of all pages on the site, or for a list of all pages edited by a specific person, or for a spreasheet showing all website users by role. I could list examples all day, the fundamental point is that govcms is extremely capable of producing powerful datasets, but fundamentally incapable of exporting this information in any meaningful way. As such, we would like the Views Data Export module to be considered for inclusion.

aleayr:
Well worth considering in my book. Coming from a corporate web team myself, I know the exact requests you're talking about. We used to manually compile and send out spreadsheets when content review time came around, this would help in that situation (and many others).

fiasco:
@srowlands @gollyg - what views export options do we have to date?

gollyg:
None at this point. The obvious choice would be [Views Data Export](https://www.drupal.org/project/views_data_export) (currently at 7.x-3.0-beta9). Another approach would be to expose it through services, but services is not currently in the distribution either, and it would need [Services Views](https://www.drupal.org/project/services_views).

WebProject2015:
While Services is not in the distribution, there does appear to be a UUID module which is in the distribution and has an unmet dependency on Services, so inclusion of Services would enable that module.  The inclusion of Services Views may also be of use in creating a data.gov.au bridge as described in issue 146

fiasco:
I've long advocated for Services module and see that as a more elegant implementation. Though it doesn't have CSV as a exportable format that Views Data Export will have.

gollyg:
The user story here mentions exporting data to be used in other ways. Services seems a good fit for this, and will easily expose methods to sort on that data etc. Obviously, the provision of json, jsonp, xml, and yaml formats supports a wide range of applications. 

Depending upon the structure of the data being requested and returned, a transformation into a CSV client side should be possible, although obviously complex nested data structures could be problematic.

But this is clearly not as easy to configure and use as the Views Data Export option in the specific case of CSVs.

More broadly, the addition of services would greatly enhance the usefulness of govCMS and its interoperability.

[...]

PR #197, opened 2016-03-12
#130 - Added the services and services views module to the make file

kurtfoster:
Added the services and services views module to the make file


fiasco:
@paulkilla in addition to adding this to the codebase, does this need to be added to the module whitelist also?

paulkilla:
@fiasco we will be having a separate task to go through all the modules currently in the distribution to create this list.  At first glance this would be a good candidate to be added to the list.

fiasco:
This is needed for an agency project so lets get it in now and we can manage those finer details later.

Issue #199, opened 2016-03-14
As a Site Builder (or Site Editor) I want an easy way to add menu items so that I can maintain the menu structure of my site.


kurtfoster:
I'm getting quite a few "that's weird" looks during govCMS training when I try and explain how to manage menus in Drupal. The standard drupal menu system doesn't provide an auto complete on the "Path" field on menu items. Users either need to remember the path (not easy) or manage the link in the node, then go back to the menu to manage position (two locations for one task). 

The [Multi-path autocomplete](Multi-path autocomplete) does provide the ability to autocomplete in the path. I think this would be a good usability improvement for govCMS.

fiasco:
I like the idea. The module architecture looks fine too, although, not actively maintained. From a technical standpoint I don't have a problem with it, although, I suggest we add it to the profie installation and provide an update hook to turn it on for all govCMS sites.

aleayr:
Yeah, this is a good suggestion. I tend to use the node, knowing the weight that I need to set things to, but that's clunky, especially if someone else has just set arbitrary weights to the menu.

Would support this.

mstrelan:
[Content menu](https://www.drupal.org/project/content_menu) is another solution which also allows the author to choose the menu item location before creating the content, or select existing content for a new menu item by searching on the node title. This would be a more significant UI change though.

fiasco:
We need to get this tested to verify its fixed, otherwise, we'll have to push back to 2.2 release.

fiasco:
Pushing back to 2.2 as this needs testing.

aleayr:
@Venkata-modgu Are you able to take a look at this and test for inclusion?

aleayr:
As per #276, pushing to next release.

Venkata-modgu:
The "Path" field is auto compleate on menu items. Tested with Site Builder and Site Editor roles.
Test passed.

PR #200, opened 2016-03-14
#199 - add the Multi-path autocomplete module including the patch at …

kurtfoster:
#199 - add the Multi-path autocomplete module including the patch at https://www.drupal.org/node/2418317


fiasco:
Thanks @kurtsalsa - if you can ensure this gets enabled on install, and provide an update hook to retrospectively enable it for existing customers, then we'll be able to commit this!

fiasco:
Thanks @kurtsalsa!

invisigoth:
@jozhao has discovered this module has an information disclosure flaw as it does not check access properly under certain circumstances. Anonymous users could get access to read information they should not have access to.

For example, it may expose a complete list of existing user accounts (user name and ID) to anonymous users: http://govcms.xing.com.au/mpac/autocomplete/menu/users

@aleayr would you like to discuss this with the security team?

pandaskii:
@aleayr cross-site request (CSRF) is also an issue

I would like to suggest a revert for further investigation

aleayr:
I've reverted this for the moment, let's see if we can get the issue fixed ASAP and get it put back in.

@kurtsalsa Is this something you need for your upcoming release specifically? Talking with @invisigoth and @jozhao, it doesn't seem a big issue to fix the issue above.

kurtfoster:
@aleayr, I don't need this module for anything, it's just a UX failing in Drupal that I've seen people are confused about when I do the govCMS training. I think it would be a good addition. I'm not sure when I'll get time to review the security issues.

kurtfoster:
@aleayr , @jozhao raised the bug on drupal.org here and created a patch https://www.drupal.org/node/2698843. I've made a slight change to the patch but I think it looks good. It's on my branch and I can create another PR if that looks acceptable.

Issue #201, opened 2016-03-15
Token Tweaks module (As a content author, I'd like my site to not show out of memory errors when editing pages that generate large token trees)


aleayr:
Some sites running on the distribution are running into issues when they accumulate large amounts of tokens. This often manifests as an out-of-memory error, which generates when the token tree being built is excessively large.

The token tweaks module modifies the behaviour of the base level token module, only showing a specific depth at any one time, limiting the memory consumed by generating large token trees.

Would be good to get this into the distro ASAP if people don't have a specific issue against including it.

fiasco:
I'm all for it, are you looking for someone to produce a PR?

aleayr:
If someone would like to, that would be great. I just don't have a machine to do that with today.

fiasco:
#202 was merged which addresses this issue.

PR #202, opened 2016-03-15
Add token tweaks module

pandaskii:
Add token tweaks module

ISSUE #201


seanhamlin:
I would be a big fan of enabling this by default for govCMS installations. Also to minimise the memory usage, you can also set `token_tree_recursion_limit` to 1.

fiasco:
I agree with @seanhamlin, @jozhao could you please add it to the installer, also provide an update hook to install on existing sites and also set `token_tree_recursion_limit` to 1. On a site-by-site basis we can increase that on the SaaS platform if needed.

pandaskii:
No problem working on it

fiasco:
@jozhao we're pretty close. Now we just need to set `token_tree_recursion_limit` which we can do on the platform.

Issue #175, opened 2016-02-23
Vimeo videos not embedding through WYSIWYG


Pixie74943:
Hey Team,

Using the WYSIWYG editor and attempting to embed a vimeo video is not working in our live site (though it was in our test site). I've tested with a wide variety of videos, and nothing from Vimeo is working. The same plugin works fine for Youtube.

Not an urgent issue as we only have a couple of videos and I'm embedding them manually via iFrames, but something to look in to if you ever have a free moment. I'm assuming the issue is with govCMS and not the module itself, as it's working on my test site.

aleayr:
Hi @Pixie74943, when you say not working, do you mean the WYSIWYG editor is getting rid of the video when you save it, or it's displaying a broken video player (like it can't load from Vimeo)?

Where is your test environment? Is it on a local network/machine, on the govCMS managed hosting environment or elsewhere?

Pixie74943:
Hi Adam,

The test environment is on my desktop. I've pushed this site up to Acquia Cloud and the videos still work nicely.

On our govCMS environment (decd-sa.govcms) we don't get any result. The iframe embeds, and takes up space on the page, but nothing loads inside of that space.

The HTML output is:
`<p><iframe src="http://player.vimeo.com/video/108854850?autoplay=1" width="400" height="225" class="video-filter video-vimeo vf-108854850" frameborder="0"></iframe></p>`

Looking at the section using the google chrome developer toolbar, nothing has been loaded inside the iframe.

Sorry my original post wasnt clear, hopefully this helps.

aleayr:
All good, I was just making sure I fully understood it.

Do you have an example URL you should give me on the decd-sa.govcms site that isn't working, so we can investigate? (and maybe an example on the Acquia Cloud that is working?)

pandaskii:
One possible reason https://developer.mozilla.org/en-US/docs/Security/Mixed_content/How_to_fix_website_with_mixed_content

Also check Google blog: https://googleonlinesecurity.blogspot.com.au/2011/06/trying-to-end-mixed-scripting.html?m=1

Pixie74943:
Hey @jozhao that looks like the issue.

Using google chromes dev toolbar, I can change the iframe source from a http:// to a https:// and it loads cleanly.

The problem then following is that I am using a https:// in the WYSIWYG editor but then when that is rendered by the rich text editor it changed it out to a http:// 

**written into the WYSIWYG**
`[video:https://vimeo.com/87523337]`

[...]

Issue #205, opened 2016-03-24
Add media support for oembed media sources


gollyg:
_As a site editor I want to manage external media within the media module so that I can better manage site assets_

Specific use cases:
- Manage a library of Vimeo videos
- Manage site specfic metadata on Vimeo videos in a re-usable way (eg tagging)
- Build lists of videos using views
- Attach Vimeo videos to content types
- Define view modes that display the video

The media module is currently part of govCMS. The [Media oEmbed module](http://drupal.org/project/media_oembed) provides support for the [oEmbed format](https://en.wikipedia.org/wiki/OEmbed), meaning that any media provider supporting oEmbed can be integrated into a govCMS site.

So whilst the above use cases refer to Vimeo specifically, any oEmbed provider could be supported.

https://github.com/govCMS/govCMS/issues/175 relates to the video filter, which is currently in govCMS. This module does not provide the rich data model around videos, reducing them to a tag in the markup that is not easily extracted.

aleayr:
I'm not opposed to this per se, I see oEmbed has a number of providers it supports by default which is good, how is this exposed in the interface from a UX perspective?

aleayr:
Following on from this, I've had a number of enthusiastic responses to 'hallway' conversations I've had concerning this, so there is a lot of support for this approach.

fiasco:
@gollyg are you happy to get a PR under way? Seems like the sort of thing we may want to have available be default too right?

gollyg:
Yep, will add that in. There was a mention of disabling the video_filter module by default. Do you want that included @fiasco ?

fiasco:
Its definately worth discussing but perhaps its a different issue.

PR #207, opened 2016-03-29
Update video_filter to 7.x-3.3

pandaskii:
This update fixes issue #175 Vimeo videos not embedding through WYSIWYG

Also following patches are resolved:
1.  [Reliance on showModalDialog breaks CKEditor plugin in Chrome](https://www.drupal.org/node/2335169)
2. [Youtube autoplay always on](https://www.drupal.org/node/1430680)
3. [Youtube does not work on HTTPs enable drupal site](https://www.drupal.org/node/2438439)


fiasco:
This makes sense to me. I don't have any issue with committing it. There is #205 which is suggesting oEmbed is a better solution which I also agree with but I don't think that should stop fixing what we currently have in govCMS.

aleayr:
This jumps a couple minor point releases, release notes for the last two versions 3.2/3.3 are pretty succinct nothing concerning. Happy to merge?

PR #208, opened 2016-03-30
Revert "#130 - Added the services and services views module to the make file"

aleayr:
Reverts govCMS/govCMS#197

More discussion required about implications of introducing the Services module to the distribution/SaaS hosting environment.


rujiali:
Hi guys,

There is requirement from client wanting to implement AJAX call within the same page. Services and Services Views module would be necessary to implement one page design user experience. I'm wondering when these two modules will be in the repo.

Thanks,

Josh

fiasco:
@rujiali we're currently tracking the use for Services and Views module in #223. If you have another use case for these modules that doesn't align with their use in #223, we'd appreciate you logging a seperate issue. This helps us better understand the need and application of these modules to govCMS.

PR #213, opened 2016-04-01
Update Date module to 7.x-2.9

mstrelan:
Many bug fixes, see https://www.drupal.org/node/2565073. Specifically for my case the bug fix required is [Date format in calendar title remains default (e.g., month view date shows as July 1, 2014)](https://www.drupal.org/node/2294973).


aleayr:
Since this is a point release, and I don't see anything in the release notes that concern me, does anyone have an issue with merging this in?

Otherwise, anyone with any suggestions on how we should approach testing this (acknowledging that it will run through our verification process with test sites on the S platform as well)?

aleayr:
This will go through the test process over the next fortnight for agencies to test their websites before full prod deploy, so merging.

Issue #119, opened 2015-12-04
As a site editor I want to create 'mini-sites' within my main site


aleayr:
An agency has listed a requirement for adding mini-sites to their govCMS site (sample file attached).

Suggested approach was to upload a zipped archive attached to a node, and automatically extract this into the files directory. When the node is deleted, it should clean up all files uploaded. If a zip file contains any not-allowed extensions, the entire upload must be aborted, and no files saved to the file system.

h3. Requirements

File extensions permitted by default:
- .js
- .css
- .htm
- .html
- .png
- .jpg
- .gif
- .pdf
- .doc
- .docx
- .ppt
- .pptx
- .xls
- .xlsx
- .tif
- .xml 
- .txt

Extensions to never be allowed:
- .exe
- .scr
- .bmp

Zip must:
- contain index.html
- not exceed a certain compression ratio (e.g. a zip bomb)
- must pass clamav scanning

h3. TODO
- What do we want to do around having no GA tracker on the pages?
- How about accessibility of the zip file sites? How is this to be handled?

aleayr:
Comment from Jane:

Further information is requested on how ClamAV handles compressed files (e.g. zip). Such files can be problematic for virus scanners and often not permitted.

aleayr:
Comment from @srowlands:

Initial code review of this module indicates it requires two additional dependencies:
    Field redirection
    Registry autoload

Registry autoload is intended to 'bridge the gap' between Drupal 8 and 7, allowing modules to be written with PSR-0 and PSR-4 coding standards. We would prefer modules abide by already existing Drupal 7 coding standards to keep the codebase consistent and recommend refactoring the files and functions within this module to remove the need for the 'registry_autoload' module.

Initial security review reveals no glaring issues, in the way the module functions generally:
- zip files containing disallowed files are not allowed
- the module cleans up temporarily extracted files immediately if they contain files in the disallowed list
  when a node is deleted the mini-site files are deleted too
- if revisions are deleted the mini-site files for that revision are deleted too
- archive files will not extract if they are in an invalid format

There is one (potentially) major issue though:
- A user with sufficient privilege can override the allowed extensions

I would recommend making this a hardcoded fixed list to prevent abuse.

aleayr:
Comment from Teresa:

Explicitly what existing govCMS role could override the allowed extensions ie site editor, site builder, content author?

aleayr:
Comment from @srowlands:

[...]

PR #217, opened 2016-04-06
Add Minisite module

aleayr:
Longstanding request for old structured content (such as annual reports that contain their own look and feel) that won't get edited to be served by govCMS.

This is a first release, and there will be updates to the module in due course to expose content to search which isn't available currently (only spider indexes like Google will index it - not the internal search).

Documentation for the module and it's use is coming shortly.


aleayr:
Solves #119.

aleayr:
I've extensively testing this through demoing to agencies and happy to merge.

PR #218, opened 2016-04-06
Added module module_permissions

aleayr:
This is the first step to allowing SaaS platform customers to be able to manage config of specifically defined modules.

This puts it into the distribution, but will be off by default, giving us the capability if required before a full launch proposed in release next + 1.


aleayr:
Security team has given approval. Will be disabled by default, and can be enabled on a case by case basis. Later, the functionality will be enabled by default.

PR #221, opened 2016-04-11
Removed ACSF module as tracked dependancy

fiasco:
As we no longer track SaaS integration dependancies (as they don't value anybody but the SaaS platform), the ACSF module is no longer needed in the distribution.


aleayr:
Confirmed this is now installed as part of the SaaS platform specific build tools. Merging.

PR #177, opened 2016-02-23
Patch Media overlay bug (beta1).

stooit:
Alternate to #133 to patch the currently running Media-beta1 with the backported patch from 2.x-dev.

This may be a more preferable option to updating to 2.x-dev and introducing new, currently unknown change.

Issue #132, opened 2015-12-16
Media browser dialog opens twice for image fields


stooit:
When using the Media Browser on an image field the overlay opens twice, and you cannot select images until you close the top overlay (an error "You have not selected anything!" is thrown).

This issue has been replicated on production sites and a vanilla local install of govCMS master. It looks like this issue was introduced recently, but I haven't dug into by what.

This is a duplicate of the issue posted on Drupal.org (https://www.drupal.org/node/2534724).

Pull request incoming to incorporate patch 53 from the d.o issue.

aleayr:
Looks like there is a more recent patch from the issue queue, patch 72, might be good to roll that in instead.

kurtfoster:
@srowlands , @aleayr , I'm still getting this issue on beta 9. We're going live very two weeks from now and content entry on our job starts tomorrow. Is this bug fix included in beta10 and if so, when can we expect a release?

stooit:
@kurtsalsa I resubmitted the PR for this a few days ago, hopefully it gets reviewed and merged soon. Has been bugging me for months!

kurtfoster:
@srowlands, ok, thanks for the info. For now I've manually added the patch to my repo so that content entry can move ahead properly this week.

Feng-Shui:
@srowlands Thanks for sorting this, applied the patch this morning and it's working nicely for us.  We're going to patch locally too until this gets merged in, been a real frustration for out editors.

andmair:
@fiasco @gollyg this issue is also a problem for NFSA, which is a very media rich site.  They will need this patch on ACFS before they go live in October.

fiasco:
#227 is the current pending PR which requires a re-roll. @andmair if we can get that re-rolled we'll be in a position technically to commit it (assuming it passes testing and gets in-roadmap).

stooit:
@fiasco re-rolled, you should be able to confirm the fix in the PR environment. It should be noted this patch has had fairly thorough testing; it's in use on many govCMS projects outside of SaaS without known issue

fiasco:
Thanks @srowlands.

@aleayr could we prioritise @Venkata-modgu to test this?

aleayr:
@Venkata-modgu Can we get this tested please?

Venkata-modgu:
Media browser interface opening only once.
Test passed.

PR #227, opened 2016-04-14
Patch media to fix double-overlay bug.

stooit:
Replace #177, patches media to fix double overlay appearing (see Issue #132)


fiasco:
@srowlands can we get a re-roll for this?

stooit:
Ha, literally had my terminal open doing that already. Will be in shortly @fiasco

Venkata-modgu:
Media browser interface opening only once.
Test passed.

PR #230, opened 2016-04-18
Update CKEditor to 4.5.9

mstrelan:


Venkata-modgu:
Test partially passed. Text formatting like color and font family are lost when using 'past from word' option, apart from that everything is good.

fiasco:
Do we let color and font-family in anyway? I would've thought we'd prevent that type of formatting from being stored with the content anyway as its more of a theme control than content.

PR #240, opened 2016-04-25
Patched Context UI module to reduce memory usage

fiasco:
Re-opening #190 to merge against new branch naming convention


fiasco:
@chris820 makes a good point in #190 that choosing to remove options from the registry has greater implications. The issue, after all, is generating massive amounts of context conditions are both time consuming and memory intensive. The better option would be to provide scalable mechanisms like ajax driven, on demand condition generation or autocomplete instead of menu based options. But these solutions belong in the Context project issue tracker and not here. That leaves us with a couple of choices:
- Use the registry alter technique as a coping mechanism on the platform. Its not a UI exposed option which buts the control into the developers hands giving us more more control around sanity checking such as ensuring we're not removing context conditions already in use.
- Do nothing and wait for a better solution from the Context project? Probably won't happen though now that the focus has shifted to Drupal 8. Unless we do something about it.

fiasco:
Given the gains this patch would represent, there isn't a good reason to commit this to govCMS. We should instead wait for a fix to the context module. SaaS platform may resolve this in other ways.

PR #241, opened 2016-04-27
Update Redirect to 7.x-1.0-rc3

mstrelan:
Redirect 7.x-1.0-rc2 includes a fix for the circular redirect patch that GovCMS is tracking, and also allows for redirects to be disabled on appropriate entity types. Few extra bug fixes in rc3.


fiasco (approved):

mgdhs:
Was the [update hook issue](https://github.com/govCMS/govCMS/issues/440#issuecomment-275963680) resolved in rc3? If not, you'll get a [PDO exception when running updb](https://www.drupal.org/node/2578019#comment-11898757). @fiasco

seanhamlin:
@mgdhs nope by the looks, we will be manually running a SQL query on the SaaS platform, see https://github.com/govCMS/govCMS/commit/2cfd9df6181a92c9e7106b0dde773855f860c66e#diff-74764508a513e1ac31b838a8c8877218

mstrelan:
Can't this be done via an update hook in govcms.install or one of the govcms modules? You can use hook_update_dependencies() to declare redirect_update_7103() is dependent on govcms_update_N().

seanhamlin:
Good question, this certainly would make the release more repeatable, and less prone to error. It will also help anyone outside the SaaS environment to do the upgrade.

One thing I see as potentially an issue, is that the update hook list has already been generated when you run `updb` the first time, so I don't think you can alter this list in flight. The first time would always error, and then the second time `updb` would succeed.

I did spot a [patch](https://www.drupal.org/node/2578019#comment-11898757) to simply rename the update hooks, perhaps this is the path of least resistance? Although, this works until the next release with an update hook comes out. And when that does, then we end up in this same pile again. 

Another option would be to add release notes to the public release of govCMS 7.x-2.13 to include something along the lines of "you need to run this database query before you upgrade".

Keen to hear thoughts.

tobybellwood:
given the ongoing conversation about this, I think the best action currently is to revert this merge and move it to the future milestone.

PR #50, opened 2015-09-22
First pass to replace iconomist and favicon with responsive_favicons

typhonius:
This addresses #31 - it's a super rough first pass that we can build on so please be gentle :kissing:

Issue #31, opened 2015-07-22
Replace iconomist and favicon with responsive_favicons


typhonius:
The iconomist and favicon modules could be replaced with [responsive_favicons](https://www.drupal.org/project/responsive_favicons). This will allow us to reduce the module footprint of govCMS whilst also increasing functionality.

This patch will need to:
- Add responsive favicons to the make file
- Add an update hook to a govCMS module (probably tweaks but perhaps the profile) to port any settings on sites which have these modules enabled.
- Disable and uninstall those modules
- Perhaps disable those modules from being enabled (hook_paranoia_disable_modules())?

We'll also need a follow-up patch for removal of those modules from the make file.

Patches welcome

fiasco:
Seems like a good idea, can we get a pull request for it?

aleayr:
As part of testing and merging this in, we need to consider how this will effect (and how we test) the change on existing sites.

fiasco:
@aleayr agreed, but we need to do that with a pull request provided ;)

fiasco:
@typhonius we're keen to see this in govCMS but we need some help:
1. we need to reroll the PR
2. we need instructions for how an agency can test this change

typhonius:
I've addressed this in #247. An agency should undergo the following steps:
- Patch govCMS codebase to include change
- Run update
- Ensure favicon and iconomist modules are disabled, uninstalled and unable to be reinstalled
- Ensure that responsive_favicon module is enabled if favicon module was enabled prior.

aleayr:
@Venkata-modgu Can we get this tested #247 for merging?

aleayr:
Due to complexities in modifying existing customer sites, and depecating this, I'm closing.

PR #247, opened 2016-05-09
Disables and uninstalls iconomist and favicon modules

typhonius:
Continuation of #31 and #50


govcms.info
```diff
@@ -145,6 +145,7 @@ dependencies[] = ccl_views
 ; Custom
 
 dependencies[] = options_flatten
+; @TODO remove this after responsive_favicons has been added.
 dependencies[] = iconomist
```
fiasco, line 149: @typhonius can we remove the dependancy now? Why would it need to stay? We'd also need to remove favicon as a dependancy too and add responsive_favicons.

wiifm69, line 149: I would imagine it is because you need to uninstall the module first. A later release would need to actually remove it from the codebase.

fiasco, line 149: Why do you need to uninstall a module to make it not a dependancy?

fiasco:
@seanhamlin is there a way to automate the upgrade path for customers to move their favicons?

seanhamlin:
Nope, each agency will need to source a 260x260 PNG (with optional transparency) favicon, and then generate the necessary code in the realfavicongenerator website, then return to Drupal to upload the code + tarball.

So yes quite manual.

If you don't do the above, then you only get the stock Drupal favicon, so not a biggie. Agencies can do the needful above when they get a chance.

fiasco:
Without this being automated, I can't see us being able to get everyone off of the older modules. To that extent, I'm also not kean to add additional modules to maintain. 

So if we can't automate, I think we shouldn't merge this PR.

typhonius:
I respectfully disagree. I understand it might be a hurdle without automation, however we're slimming down our module count, moving towards contrib rather than sandbox and fixing bugs by removing.

Absolutely there should be agency consultation and engagement but for fear of sounding too much like George Orwell, this sort of change is for the greater good of the platform.

aleayr:
Closing due to complexity in rolling this out. Will revisit at a later stage if required.

Issue #271, opened 2016-06-16
Drupal 7.43 regression: When an anonymous user submits a form with an un-uploaded file that leads to a validation error, the file is lost on the next correct submission


pandaskii:
Drupal 7.43 [Release notes](https://www.drupal.org/project/drupal/releases/7.43)
When an anonymous user submits a form with an un-uploaded file that leads to a validation error, the file is lost on the next correct submission

To reproduce using the Standard installation profile:
1. Allow anonymous users to create article nodes.
2. As an anonymous user, go to the node creation page. Leave the article title blank. Select a file for the Image field, but do not click "Upload".
3. Click "Save". There will be a validation error because the title was not filled out. If you scroll down to the Image field, you'll see that the file was uploaded.
4. Fill out the title and click "Save" again. The node will be saved, but the file you uploaded before will no longer be attached to it.

This effects current govCMS sites which allow anonymous user submits a form with attachment. We have to apply a hot fix patch regarding this issue - it is in [Drupal offical 7.x line](http://cgit.drupalcode.org/drupal/commit/?h=7.x&id=6b6bc1b7fd69de28c490c976101d255b3f79d409) but labeled as [7.50](https://www.drupal.org/node/2678822) release

Feng-Shui:
Thanks for getting onto this so quickly and after hours.

invisigoth:
Just to clarify, the file may still be uploaded to the server but the webform may fail to associate the uploaded file with the corresponding submission. This issue can be reproduced by either attaching a file to a form using the "Upload" button next to the file field; or submit the form with invalid inputs directly, which triggers a form validation error. 

Symptoms of this issue have been reported to us by govCMS users.

fiasco:
@Feng-Shui if you can confirm the patch fixes the problem, we'll commit to HEAD so users can apply fix from source.

Do we need to roll this out for SaaS sooner than the 7.x-2.4 milestone? If so, we'll need to hotfix.

pandaskii:
A local test to reproduce the steps from @invisigoth also suggested the file will be deleted after 6 hours. 

As we are facing loss data, it is better to apply a hotfix after testing.

fiasco:
@jozhao still awaiting a test to verify the fix - if you can do that, I'll get this into govCMS head.

pandaskii:
@fiasco I've tested the followings in my local environment
[...]

PR #272, opened 2016-06-16
#271 Drupal 7.43 regression - govCMS 2.x

pandaskii:
Issue #271 

This patch is in Drupal core 7.x line but labeled as 7.50.


fiasco:
7.50 might be a while away. I'm happy to track this in govCMS.

PR #274, opened 2016-06-16
Update minisite module to 1.0-alpha1

pandaskii:
Release notes
- AddMinisite URL alias (experimental)
- Optionally use current page URL (defined in URL path settings) as minisite base URL


Venkata-modgu:
Minisite module updated to 1.0-alpha1.
Site editor is able to use minisite field with existing content type and new content type.
Minisite field is able to be used with page URL and URL alias. 
Test passed.

PR #229, opened 2016-04-14
Add Consultation feature.

stooit:
Resubmission against `7.x-2.x` of the original #150 
## Brief:

The Consultation feature enables a new content type (Consultation) that allow for public input via formal and informal means against a given topic.
## Functional specs:
- Ability to configure components of Consultation feature
  - Optionally enable Twitter integration (account details required)
  - Optionally enable Disqus commenting (account details required)
  - Optionally enable private submissions (requires opt in)
- Ability to create new consultations, which include the following:
  - Basic details (Title, Summary, Body)
  - Start/End date for a Consultation
  - Related issue documentation (file attachments)
  - Outcomes of a given Consultation
  - Updates/Milestones as a Consultation progresses through lifecycle
  - Ability to enable formal submissions
  - Ability to enable informal submissions
  - Ability to enable accepting late approvals via special URL
  - Ability to specify who should receive email notifications of new submissions
- Formal submissions
  - Can be anonymous (name will not display)
  - Could be private (entire submission will never display)
  - Could be a comment/text only
  - Could include file attachments (.pdf, .doc, .docx by default, configurable)
  - Can be made by individuals or on behalf of a company
  - Each need to be individually approved within govCMS before they will display publicly
  - Approved submissions (that weren’t marked as private) will display against each Consultation
- Informal submissions
  - Twitter discussions
    - manually curated through https://curator.twitter.com/
    - display using Twitters new grid display
  - Disqus comments per Consultation
#### Note:

This module brings a new dependency: [UUID Features](https://www.drupal.org/project/uuid_features). This is required for the Consultation Webform and included taxonomy terms. It's the successor to a module we are already using (Default content) and provides support for a wider range of entities in Drupal.
If this PR is accepted we could look to move the example content modules (govcms_content_*) to use uuid_features instead.
## DEMO:
- [Example completed Consultation](http://150.govcms.xyz/final-consultation-nbn-non-commercial-services-funding-options)
  - as originally seen on [The Department of Communications and the Arts](https://www.communications.gov.au/have-your-say/final-consultation-nbn-non-commercial-services-funding-options)
  - this Consultation has also been configured to allow late submissions via a [special URL](http://150.govcms.xyz/final-consultation-nbn-non-commercial-services-funding-options?fso=2788221432).
- [Example open Consultation accepting formal comments and files](http://150.govcms.xyz/example-consultation-accepting-formal-comments)
  - You can configure to accept either or both
[...]

PR #283, opened 2016-06-29
Add Consultation Module

aleayr:
This builds off the work from @srowlands in #229.

I've simply added new royalty free images as requested before launch.


aleayr:
The images I've places in here, I've made exactly the same size as the old ones. All good for merging.

Issue #295, opened 2016-07-07
[META] Update to Drupal 7.50


fiasco:
This is a meta ticket to track the merging Drupal 7.50 into govCMS. This update is rather large and has 2 key changes we need to address in govCMS:
### New "administer fields" permission added for trusted users

We need to check and modify the existing roles in govCMS to ensure Admin and Site Builder roles can continue to use the field UI as the have done in the past.
### Allow search engines access to CSS, JavaScript and image files

It is possible to write an upgrade hook in the distribution that attempts to update the robotstxt module configuration to insert this update into an existing sites robots.txt output so that govCMS sites can inherit this change.
### Advisory

In addition to the two above changes, there is an additional change in 7.50 that may break existing sites:
#### Drupal 7 core is now protected against clickjacking by default.

If a site is using the Security Kit module that already writes the X-Frame-Options header on its own, that setting will be automatically respected and Drupal core will not overwrite it. For all other sites, the X-Frame-Options header will be emitted and any uses of govCMS sites as embedded content will cease to work. However, once reported, the seckit module can be enabled and configured to allow specific domains to embed govCMS sites.

SaaS users will be able to file Support tickets to address this issue. All other uses of govCMS will need to address this themselves.

More Information: https://groups.drupal.org/node/512445
Community post about clickjacking: https://community.govcms.gov.au/t/updates-coming-to-govcms-that-may-break-your-site/248
### UTF8MB4 support

This essentially gives Drupal the ability to store 4 byte characters in the database. Emoji characters are typically the main reason why people want this supported, but sometimes extended Chinese characters are also needed.

**N.B.** By updating to 7.50 will not automatically grant you 4 byte character support, you also need MySQL configuration changes, settings.php changes and your database needs converting. More information can be found in the conversion tool https://www.drupal.org/project/utf8mb4_convert

rooby:
This is great news. The administer fields permission means that site builder will now be able to access all the field UI pages.

I'd be happy to help out with whatever is necessary to get this through as quick as possible so let me know if I can assist in any way.

fiasco:
Thanks @rooby! As you can see, we've targeted this change for the 7.x-2.6 release which is [scheduled](https://github.com/govCMS/govCMS/wiki/govCMS-Release-Schedule-and-Roadmap) for the for [28th of July](https://github.com/govCMS/govCMS/milestones).

In order for that to happen, we need 3 PRs:
[...]

PR #296, opened 2016-07-09
Drupal core 7.50 update

rooby:
For issue #295
- Update Drupal core to 7.50
- Remove patch for https://www.drupal.org/node/2678822

### DIFF
--- a/drupal-org-core.make
+++ b/drupal-org-core.make
@@ -1,7 +1,6 @@
 core = 7.x
 api = 2
-projects[drupal][version] = "7.44"
+projects[drupal][version] = "7.50"
 projects[drupal][patch][] = https://www.drupal.org/files/issues/drupal-7.x-allow_profile_change_sys_req-1772316-28.patch
 projects[drupal][patch][] = https://www.drupal.org/files/issues/drupal-1470656-26.patch
 projects[drupal][patch][] = https://www.drupal.org/files/issues/core-111702-99-use_replyto.patch
-projects[drupal][patch][] = https://www.drupal.org/files/issues/drupal-n2678822-22.patch


fiasco:
test this

seanhamlin:
Looking at the automated site install logs in Jenkins

```
The following module is missing from the file system: content. In        [error]
order to fix this, put the module back in its original location. For
more information, see the documentation page. in
_drupal_trigger_error_with_delayed_logging() (line 1128 of
/var/lib/jenkins/jobs/govcms_pull_request_builder/workspace/296-actual/docroot/includes/bootstrap.inc).
[63.22 sec, 96.96 MB]
```

Some quick googling found https://www.drupal.org/node/2487215 and also this blog post http://www.vdmi.nl/blog/i-went-drupal-733-and-all-i-got-was-slow-site

So it looks like either a contrib module or custom one is trying to load a file that does not exist.

I would recommend debug backtracing a local site install and seeing what module is causing this. This PR might be blocked until this is sorted.

seanhamlin:
Also someone wrote this module which may help https://www.drupal.org/project/module_missing_message_fixer

seanhamlin:
This is an example issue (module not in govCMS but it gives you an idea) https://www.drupal.org/node/2763555 to show you what the bug is.

seanhamlin:
Using some grep foo, I think I have found it 

```
$ grep -nriI "module_load_include('inc', 'content'" *
profiles/govcms/modules/contrib/google_analytics_reports/google_analytics_reports.install:97:    module_load_include('inc', 'content', 'includes/module');
```

There is an issue for google_analytics_reports - https://www.drupal.org/node/2762675, and it seems that 7.x-3.1 includes this fix.

seanhamlin:
The more I think about this PR, I think you should just push the update to google_analytics_reports in this PR, as we need to bump them in tandem.

pandaskii:
Agree with @seanhamlin we have two options here:
1. A PR to bump google_analytics_reports, then rebase this PR
2. In this PR, include google_analytics_reports 3.1

More thoughts, should we remove dependencies[] = google_analytics_reports from govcms.info?

seanhamlin:
My vote is for option #2, in a squashed single commit, with appropriate messaging to indicate why they are included together. This makes the most sense to me.

As for removing the google_analytics_reports from the default install, I am unsure on what the google_analytics_reports module even does, but that is certainly worth a follow up issue. I was under the impression that all agencies get their own analytics reports from GA directly (and not from within Drupal).

aleayr:
Hi @rooby,

Would you be able to perform Option #2 above with this PR (including google_analytics_reports version 3.1)?

Then we can finalise testing and roll this in. Let me know if this presents an issue.

fiasco:
test this

aleayr:
Closed in favour of #368.

Issue #300, opened 2016-07-13
Consultation module review


Feng-Shui:
Morning all,

I've been asked to do a review of the consultation module, we're interested in using it at some point.  My comments are below.  I'm happy to open pull requests for any of these.

Thanks for the work on this module!

**1: Undefined constant COMMENT_NODE_OPEN**

This constant is declared by the comment module and is used when preprocessing consultation nodes.  The comment module is not listed as a dependency and so this constant can undefined.  It doesn't cause any functional issues, but the code should check if it's defined before comparing against it.

**2: Code comments**

`govcms_consultation_theme_registry_alter()` is missing a doc block and a number of the helper functions don't document return types etc.  I'm happy to document these.

**3: Un-used code**

`govcms_consultation_form_node_form_alter()` has an if condition in it with no code inside, and there are a few un-used variables in `govcms_consultation_preprocess_node()`: `$end_consultation_date` and `$time`.

**4: Consultation duration not calculated correctly**

`_consultation_vars()` is calculating the duration of a consultation incorrectly.  It’s doing a subtraction between two timestamps which is problematic due to timezones and not being inclusive of the end date:

```
1451599200 =  1 Jan @ 9:00am
1454220000 = 31 Jan @ 5:00pm

(1454220000 - 1451599200) / 86400 = 30.3 = 30 when rounded
```

Suggest that this is replace with a date based comparison which ignores time.

**5: Namespace for fields and functions**

There are a few helper functions which do not have the module's namespace prefix, `_consultation_vars()`, `_consultation_percentage()` and `_consultation_days_remain()`.  These should be namespaced inline with the modules project name.

The module also doesn't namespace it's declared fields, and they include some quite generic field names which could foreseeably be used by other modules or site builders: field_icon, field_file_word, field_updates etc.

However, I note that this occurs across a lot of the features, with govcms_events declaring field_cost, field_contact and field_location, govcms_publications declaring field_date etc.  Prob not much can be done about this as this stage.

**6: Entropy in `_govcms_consultation_admin_return_salted_hash()`**

My crypto foo is totally rubbish, but thought it might be worth mentioning that this function doesn't use a lot of entropy when seeding `crc32()`.  It passes in a hash set at install time `drupal_hash_base64(uniqid())` and then the node ID which is available from the source.  Two hashes for different nodes might only have a variation in their seed data of 1 character.

No idea if that's an issue, but if so it could be seeded with the modules UUID, or throwing in another random string etc.

[...]

PR #305, opened 2016-07-19
#300 - Consultation module tweaks

Feng-Shui:
This PR contains the following changes to the consultation module:
1. Ensure COMMENT_NODE_OPEN is defined before being used for evaluation
2. Remove used variables in govcms_consultation_preprocess_node()
3. Add module namespace to helper functions
4. Remove return from _govcms_consultation_consultation_vars(), $variables param is passed by reference
5. Set default return value of _govcms_consultation_admin_return_end_consultation_date() to be an int as it’s used in numeric comparison
6. Seed the late submission URL token with the node UUID rather than NID.
7. Refactor _govcms_consultation_admin_return_end_consultation_date() to simplify parameters, `$node` no longer required.
8. Add/update function doc blocks.
9. Remove global $base_url reference in _govcms_consultation_admin_return_late_submission_url() and let url() function handle absolute path.
10. Namespace module constants.


fiasco:
Would have prefered a few PRs for all of this as there is quite a bit to review and no automated tests to ensure nothing has broken in this change. Will need someone to login into http://305.govcms.xyz/ to test.

Feng-Shui:
If you take a look at the diff,  it's pretty straight forward, most of these commits are one liners and a lot of the changes are actually comments.  You're preaching to the choir re test coverage, I'm adding some in now that I'm working on the functional changes.

fiasco:
I've done some initial testing on the test environment and it doesn't appear to break anything so I'm going to merge this one as its primarily clean-up.

Feng-Shui:
Awesome, thanks @fiasco being able to re-branch from this will make the remainder of the changes much easier.

PR #306, opened 2016-07-20
Enable govCMS account security module by default

pandaskii:
About half a dozen agencies in the first half of this year have reported account lockout issues with GovCMS, which have been successfully resolved by enabling the “Account Security” module. It is about time to enable this module by default so agencies with accounts disabled by the default security settings in GovCMS can resolve those issues by requesting password reset emails.

### DIFF
--- a/govcms.info
+++ b/govcms.info
@@ -152,6 +152,7 @@ dependencies[] = user_action_report
 
 ; govCMS
 
+dependencies[] = govcms_account_security
 dependencies[] = govcms_appearance
 dependencies[] = govcms_core
 dependencies[] = govcms_common_fields


xtfer:
I left a note on #194 that is related to this. The use of hook_init() in this module is a bit problematic.

aleayr:
Due to other modules like r4032login, login redirect and search404 firing very early, this is the only way to solve this issue. Longer term, there have been discussions about reducing the use/dependency on the other modules above that resulted in this outcome.

Venkata-modgu:
govCMS account securitay module enabled by deafult.
Test passed.

PR #308, opened 2016-07-21
#307 Remove old tests

### LINKED ISSUE #307: Profile testing files missing/out of date
Morning all,

I'm working on some test coverage for the consultation module, and was going to add the new consultation node type to the profile's basic test coverage.  However, the test files to not exist in the profile directory (not copied in as part of the build process), and when copied in manually, they don't run anyway because they appear out of date.

The files are:

```
build/tests/govcms_test_base.inc
build/tests/govcms.test
```

They are referenced in the profiles info file

```
files[] = tests/govcms_test_base.inc
files[] = tests/govcms.test
files[] = tests/govcms_metadata.test
```

Note that `tests/govcms_metadata.test` is missing completely.

When copied in manually and run, the tests batch fails with fatal errors `Call to undefined function govcms_supported_themes()` and the content type features referenced in the `setUp()` don't appear to exist in govCMS anymore.

If this was an intentional change I'll open a PR to remove these, if not we should get these up to date - or have we replaced them with behat testing now?

Feng-Shui:
Resolve #307 - removing legacy simpletests.


fiasco:
Too easy. As the priority is for consultations to get into 2.5 I won't merge this until after 2.5 is cut.

PR #313, opened 2016-08-03
Update govcms_ckan to latest stable release.

stooit:
screen shot 2016-08-03 at 5 01 54 pm
## Update to latest (v1.1) stable release.

govCMS CKAN v1.1 includes many bug-fixes and improvements, including:
- Advanced tick options
- Advanced label options
- Advanced key override options
- Advanced axis value options (format + limit)
- Ability to draw lines + labels at arbitrary points on a graph
- Fix some chart settings not being set correctly
- Override legend and axis labels
- New visualisation: Scatter plot
- New inline help
- Many bug-fixes


fiasco:
we'd love to see some behat tests here that test/prove use cases for this module in govCMS

stooit:
No problem, but can the behat tests come in a follow-up PR? There will be some complexity.

fiasco:
This update won't get in without some form of testing being completed. @Venkata-modgu is our only testing resource at the moment at there is a huge backlog to work through. This change doesn't currently have any advocacy or priority for manual testing.

So automated testing will be the best way to fast track its adoption. Granted you may need to find a way to run behat features tracked in this module rather than features directly in govCMS.

gavintapp:
Adding a note regarding this issue. We are actively working on a govCMS project that relies on the newer version of this module. We will create some behat tests in the next few days. Can we have an indication of where this issues sits in the roadmap? It would be good to have it included in the distro before 27/10/16 when this module is showcased at Drupal South.

seanhamlin:
Here is a link to the changes this introduces https://github.com/govCMS/govcms-ckan/compare/7.x-1.0...7.x-1.1

One question I have is that will existing sites have an upgrade path from 1.0 to 1.1? I did not see any update functions, so I am assuming all existing graphs created under 1.0 will continue to work?

gavintapp:
Yes, sites created using version 1.0 will continue to work with 1.1 

In addition, a site on govCMS that was migrated from dedicated to shared hosting will see several issues fixed by this update, as it was developed using 1.1

seanhamlin:
I have read through the extensive git file changes, and am of the opinion we should merge this now, with the assumption that a follow up PR will be included to visit some pages created with behat that include CKAN graphs on them.

Manual testing can be done for all sites that use 1.0 on govCMS SaaS to ensure there are no regressions (on the TEST environment).

@aleayr I will leave this one in your pile, but I believe we need an answer fairly quickly here.

aleayr (approved):
Need to deploy to staging to ensure this doesn't break backward compatibility with other CKAN functionality from the 1.0 release.

Issue #303, opened 2016-07-15
A pipe separator appears in html title tag on govCMS site front page


pandaskii:
A pipe separator appears in html title tag on govCMS site front page - this is happened after new govCMS release

After investigation, this is from a module - metatag's default settings.

Feng-Shui:
https://github.com/govCMS/govCMS/pull/316 resolves this issue, test to verify the issue and then the fix.

fiasco:
Merged #316 which addresses the issue in future releases but need to look at whether this is still an issue for existing sites.

fribibb:
Using https://gist.github.com/fribibb/e0844a0dc53bb474f122895308a865ab and running:
`casperjs all-sites-check.js | grep "|$"`
the following 18, live govCMS sites have a trailing `|` in the `<title>` tag:
- Tax Practitioners Board
- ATO Beta
- Australian Training Awards
- Unique Student Identifier
- Australian Apprenticeships
- Professional Standards Board
- Shared Services Centre
- Department of Employment - Strategic Plan 2014
- Department of Employment
- IP Australia
- Australian Sports Anti-doping Authority - ASADA
- Office of National Assessments
- Commonwealth Director of Public Prosecutions (CDPP)
- Office of the Registrar of Indigenous Corporations
- Victorian Budget 2016/17
- Election Costings 2016
- ATO Annual Report 2014-15
- Australian National Audit Office

I'll script up something to check and fix all existing sites - so it will fix live as well as offline sites in blue/green environments.

marblegravy:
Just coming in to this issue seeing a bunch of the sites we look after in the list above - what's the issue with the pipe separator?
Is it just concerns over how some screen readers read out the pipe as "vertical line"?

aleayr:
It's a bug introduced on the homepage of sites in a previous release where the site title will be trailed by a pipe character even though there isn't a second textual component.

Eg, front page title of ANAO is: "Australian National Audit Office |" -> should be "Australian National Audit Office"

aleayr:
I should note there isn't anything wrong with the sites operationally, it's just an aesthetics bug we'd like to fix up.

aleayr:
Was this resolved @govCMS/osb-devs?

I know @courdo1 and/or @fribibb were looking into this.

fribibb:
No sorry @aleayr, both @courdo1 and I have been pulled onto other things. I'll try and make some time next week for this.

aleayr:
[...]

PR #316, opened 2016-08-05
Feature/303 metatag front title pipe

Feng-Shui:
Resolution for #303

This is an initial test to confirm the bug, it should fail...


Feng-Shui:
Went for a function in profile install rather than rolling a config feature to set one variable - seemed like overkill.

fiasco:
Given there are tests for this that clearly fix the problem, I don't have an issue in merging this. @aleayr can you confirm?

DJMyles:
Verified this bug is no longer present on a fresh install

Issue #322, opened 2016-08-11
Misleading help text on consultation web form 


Venkata-modgu:
The help text provided for phone number on consultation form is misleading.
Mobile number in Australia has 10 numbers, but there are 11 numbers in the example text (0414 1234 567).

![phno-2](https://cloud.githubusercontent.com/assets/16928665/17579614/bc2a47d2-5fda-11e6-92de-2e4f5a8e1525.jpg)

PR #323, opened 2016-08-11
#322 Fix incorrect mobile number example in consultation webform.

Feng-Shui:
Simple fix for #322


fiasco:
Tested and verified it fixes issue. Thanks for the contribution!

PR #331, opened 2016-08-24
Fix Panels bugs introduced by Panels 3.6

aleayr:
In releasing the security update for Panels 3.6, the Panels module has broken in place editing functionality for Site Builders.

Panels has released the 3.7 patch to resolve this.


aleayr:
Testing passed.

PR #338, opened 2016-09-01
Add in build:no-clean task, and build:no-clean:install

paulkilla:
These will allow us to run a local build without blowing away local changes.

This is currently only possible if also running tests... Sometimes we don't want to run the tests.

### DIFF
--- a/build/phing/build.xml
+++ b/build/phing/build.xml
@@ -113,6 +113,14 @@
             description="Build the project and install govCMS."
             depends="build, install"/>
 
+    <target name="build:no-clean"
+            description="Build the project without removing local changes, install govCMS."
+            depends="make"/>
+
+    <target name="build:no-clean:install"
+            description="Build the project without removing local changes, install govCMS."
+            depends="make, install"/>
+
     <target name="build:with-validation"
             description="Build the project, install, then run code validation."
             depends="build:install, validate:all"/>


build/phing/build.xml
```diff
@@ -113,6 +113,14 @@
             description="Build the project and install govCMS."
             depends="build, install"/>
 
+    <target name="build:no-clean"
```
fiasco, line 116: This is just an alias of `make` so do we really need it?

build/phing/build.xml
```diff
@@ -113,6 +113,14 @@
...
+            description="Build the project without removing local changes, install govCMS."
+            depends="make"/>
+
+    <target name="build:no-clean:install"
```
fiasco, line 120: I'm hesitant about allowing an unclean build as it may create bugs when used by the wider community. I presume the justification here is so that changes can be built and tested before being committed? Instead, you can use git branching to do have the same effect:
1. Checkout a new working branch
2. Make developing and debugging changes and commit them to the branch
3. Once you have reached a stable place, create a new branch from the same starting point and conduct a squash commit merge of your working branch. 

In this way, you only get the release history of the single squash merge which is better for the mainline of the govCMS master branch.

fiasco (commented):
Overall I don't think these changes are necessary and may even introduce further complexity and confusion to contributors.

deanvucic:
This may be considered for future milestones. For now, this PR is closed.

Issue #339, opened 2016-09-01
Automated tests fail because referenced file “custom.files” has been deleted


pandaskii:
Error message:

Execution of target "validate:phpcs" failed for the following reason: 
"build/phing/build.xml:272:8: Reference custom.files not found."

aleayr:
This is now solved.

PR #340, opened 2016-09-01
Fix ISSUE #339 Automated test is failed for reference custom.files not found

pandaskii:
Fix ISSUE #339 Automated test is failed for reference custom.files not found

### DIFF
--- a/build/phing/build.xml
+++ b/build/phing/build.xml
@@ -80,15 +80,15 @@
     <property name="behat.local"
               value="${repo.root}/build/tests/behat/behat.local.yml"/>
     <property name="phantomjs.bin"
-              value="${composer.bin}/phantomjs" />
-    <property name="phantomjs.webdriver" value="4444" />
+              value="${composer.bin}/phantomjs"/>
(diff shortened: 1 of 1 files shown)


build/phing/build.xml
```diff
@@ -273,8 +283,8 @@
                 standard="${repo.root}/build/vendor/drupal/coder/coder_sniffer/Drupal/"
                 showSniffs="true"
                 showWarnings="true"
-                haltonerror="true"
```
fiasco, line 276: This is not a fix, this would just be ignoring the errors no?

fiasco (commented):
Can you provide more info about when this fails? AFAIK the jenkins bot that does the building doesn't seem to run into this issue (I thought)

build/phing/build.xml
aleayr: @jozhao Is this still required? When does this failure happen?

pandaskii: @aleayr I cannot test it as the build is failed every time in my local environment

aleayr (approved):
Confirmed this resolves an issue in the local build process.

Issue #302, opened 2016-07-15
Allow Site Roles to Suspend Accounts Only (remove option to cancel).


pandaskii:
Delete account and make content belong to "Anonymous" using Drupal module Views Bulk Operations (VBO) leads to lose contents in current govCMS sites. At lease one govCMS site has reported that losing data because of this bug.

To reproduce this bug:
1. On a VBO view that lists users - select one account, and then click the Delete button.
2. Select "Delete account and make content belong to "Anonymous"
3. At the end of the batch operation an error appeared and all of the user nodes are deleted

After investigation, this is caused by a bug in Drupal module Views Bulk Operations (VBO) - https://www.drupal.org/project/views_bulk_operations

Solution suggestions:
1. Disable this bulk cancel account action from VBO views. User still can be canceled/deleted in same user administration page
2. Apply a patch to VBO and we are tracking a discussion on Drupal community https://www.drupal.org/node/2505457

fiasco:
As a rule-of-thumb, I don't agree that content or users should be delete-able in government sites. I'd propose the first remediation is to remove the ability to delete a user and only allow disabling the account instead.

aleayr:
Noting this is a high priority for change.

fiasco:
The change here should be to update the VBO View to use Suspend Account rather than Cancel Account. This should solve the problem.

invisigoth:
We have confirmed this occurs when there are more than 10 nodes belonging to a user being cancelled. Under this condition function node_mass_update() in the node module tries to update those nodes in a batch job and will delete the content rather than anonymising it. The following error will be displayed when this happens:

> Fatal error: __clone method called on non-object in /.../node.admin.inc on line 323

The operation completes however successfully when there are less than or equal to 10 nodes belong to a user being cancelled.

Unfortunately the patch provided in https://www.drupal.org/node/2505457 does not fix the issue and nodes were still deleted.

@aleayr please advise govCMS users before we release the change to disallow account cancellation they should stop using the cancel user action as this will likely lead to loss of content.

Venkata-modgu:
Test passed.
Delete account option has been taken out of the cancle user account view. 
Administrator option has been taken out of the change user roles view.

aleayr:
Merged, closing.

PR #346, opened 2016-09-05
Fix issue described in https://github.com/govCMS/govCMS/issues/302

invisigoth:
Fix issue described in https://github.com/govCMS/govCMS/issues/302. 
Remove 404 from the available options when a user profile page is requested. 
Allow users with administrator user permission to review password policy settings.


invisigoth:
To test this, login as administrator:
(1) Navigate to the administer all users page and:
- try cancelling a user account. Removing the user should NOT be one of the options and cancelled user accounts can only be blocked.
- try changing roles for selected users, "administrator" should not be one of the roles that can be assigned to those users 
  (2) Navigate to configuration > review account security settings. Those with "administer users" permission should be able to see the active password policy and related settings (but cannot make changes).

Venkata-modgu:
Test passed.
Delete account option has been taken out of the cancle user account view. 
Administrator option has been taken out of the change user roles view.
Administrators are able to review password policy settings.

Issue #293, opened 2016-07-05
As a content reviewer, I'd like to view and make comments on workbench revisions that are seen in the 'My Workbench' section


aleayr:
Currently, content authors can leave comments on revisions, however, for reviewers, these can be hard to see and respond to easily.

I'd also like a solution where the revision comment is displayed in the 'My Workbench' section, and sent in emails as per #279 as well.

Podgkin:
Comments can be displayed in 'My workbench' section by creating a page clone of admin/content/comment, setting its page path to admin/workbench/content and its menu to Tab:Comments.  You can use rules to send an email to the page author or the author of a parent comment, but I don't know of a way to send it to the reviewer.  I suppose you could use an RSS feed of the comment view as suggested here https://github.com/govCMS/govCMS/issues/244 and have the reviewer subscribe to the feed.

samChinACT:
Module workbench_moderation_notes can be used to display revision comments in the 'My Workbench' section. But this doesn't create a token for the notes to be included in an email. PR for module created.

Venkata-modgu:
Test failed due to module not enabled by default.

aleayr:
Retested, with module enabled, this passes.

@samChinACT Can you update #350 to turn on workbench_moderation_notes by default?

Thanks
Adam.

Issue #153, opened 2016-01-27
address and pre tags being removed by wysiwyg filter


dylan-oxide:
The styles address and format are not retained after the page is saved. They use the `address` and `pre` tags which are removed by the wysiwyg filter.

I would suggest adding these to the settings
`address,pre`

Alternatively these styles could be removed from the styles dropdown if they are not supported.

fiasco:
This seems like a no brainer fix to implement. address and pre tags are harmless, there isn't any reason why we wouldn't support them.

teamglenny:
Agreed.

nathan-w:
Id like to see some priority on this given it's a no-brainer and its been hanging around for a while. Would be sensible to give our filter settings a sanity check in case there are a few other no-brainer tweaks we can do.

fiasco:
@dylan-oxide if you're able to put together a PR for this, we've ear-marked it for the 2.7 release.

Venkata-modgu:
Address and pre tags are not been removed by the wysiwyg filter.
Test passed.

Issue #152, opened 2016-01-27
Embedded video and sound via media are being removed


dylan-oxide:
If you use the media browser to embed video or audio sources, the WYSIWYG filter will strip these tags out. I would suggest adding the following to the WYSIWYG Filter options:

`video[controls|data-fid|data-media-element],source[src|type|data-cke-saved-src],audio[controls|data-fid|data-media-element],`

aleayr:
Thanks Dylan, good to know. I imagine when this happens you need to switch back to HTML mode? Then if a content editor switches back for subsequent edits that they disappear again?

dylan-oxide:
Correct, if you use the Full HTML text format, the markup is preserved. However if you use Rich text, the markup is removed.

For some more information, this is what is output by the WYSIWYG:

**Video**
`<video controls="controls" data-fid="7" data-media-element="1" class="media-element file-default"><source src="video.mp4" type="video/mp4"></video>
`

**Audio**
`<audio controls="controls" data-fid="8" data-media-element="1" class="media-element file-default"><source src="audio.mp3" type="audio/mpeg"></audio>
`

aleayr:
Cheers. I'm wondering whether this is the same issue as #175, in which case we could kill two birds with one stone and get this resolved. I'll confirm when Pixie74943 replies in the other issue, and get a PR in place.

aleayr:
Looks like a different issue, but I agree with this getting fixed. No use having a media browser if you can't insert all the media :)

nathan-w:
Can we package this one up with #153 ?

fiasco:
If they're different issues then we should have different PRs. But I've put them in the same milestone if we can get PRs in-time for them (👉  @dylan-oxide 😜 )

samChinACT:
fixed in #337

Venkata-modgu:
Audio and video tags are not been removed by the wysiwyg filter.
Test passed.

PR #350, opened 2016-09-06
add tags to wysiwyg, add module workbench_moderation_notes

samChinACT:
Fix issue #152
Fix issue #153 
Fix issue #293


nathan-w:
Ping: @aleayr 

Can this be split into 2x PRs please - the fix for #293 - adding in a new module is different to config change for the wysiwyg filter.

Also the wysiwyg filter needs a clean up - there are duplicated items (video, address) and the formatting of the code makes it hard to read. We should also check that the allowed attributes cover everything we want to support - eg, if Im reading the code right, the <span> tag is allowed, but any class attribute would be stripped out.

nathan-w:
Re previous comment #475 flags the need to add PRE tag to the filter. Suggest a small spike is done on the WYSIWIG so we can do a single update.

deanvucic:
This will be considered for future milestones. For now, I'm closing this PR.

PR #352, opened 2016-09-06
Update file_entity module

aleayr:
Resolves permission bug, where even though the permission is granted to Site Builders, they can't configure/administer the module.


Venkata-modgu:
Site Builder able to configure/administer the module.
Test passed.

Issue #351, opened 2016-09-06
User cannot access to Manage Fields and Manage Display tabs on file types with permission "administer file type"


pandaskii:
User cannot access to Manage Fields and Manage Display tabs on file types with permission "administer file type"

@ruwanl 

![snip20160906_1](https://cloud.githubusercontent.com/assets/1326673/18262464/0c8e14f0-7445-11e6-8823-ec57c04a016a.png)

![snip20160906_2](https://cloud.githubusercontent.com/assets/1326673/18262463/0c8d8d0a-7445-11e6-9593-97df401e2de7.png)

pandaskii:
Close the issue as the fix will be in next release

PR #353, opened 2016-09-06
Fix ISSUE #351 - User cannot access to Manage Fields and Manage Display tabs on file types with permission "administer file type" 

pandaskii:
Fix issue #351 

> User cannot access to Manage Fields and Manage Display tabs on file types with permission "administer file type"


pandaskii:
Close this one to respect #352

PR #365, opened 2016-09-23
Access permissions are bypassed when using Views Bulk Operations to change user roles.

barbun:
The original issue is here:
https://www.drupal.org/node/2516688

In a nutshell, If a user has access to a user account listing view with VBO enabled (such as admin/people when the administration_views module is used), they will be able to edit their own account and give themselves a higher role (such as "administrator") even if they don't have the "'administer users" permission.

The latest VBO module version has mitigated this vulnerability by the fact that an attacker must have access to such a user listing page and that the bulk operation for changing Roles is enabled. 

**However**, this is the case in the current govCMS where "Site Builder" and "Site Editor" roles are able to exploit this vulnerability and become administrators, which in my understanding is against govCMS playing rules. 

To reproduce: As a Site Builder or Site Editor navigate to /admin/people and use VBO at the bottom and select either "Change user roles" or "Change value" operations. You will be presented with the ability to assign any available role including "administrator". In other words, when using VBO it neglects any "Role Delegation" permissions as well.

You might think that this has to be fixed in the VBO module but the problem is actually born within the combination of "Views Bulk Operations" + "Administration views" + "Role Delegation" contributed modules. 

Proposed solution:
Alter the VBO forms and add extra permissions check to exclude the roles that the user doesn't have access to assign. Such check should not interfere with future updates of any of the mentioned contrib modules.


paulkilla:
@barbun we have resolved this as part of this fix.
https://github.com/govCMS/govCMS/pull/346

barbun:
Ah, great! Thanks @paulkilla 
Closing this PR

PR #367, opened 2016-09-29
Add Adminimal - Responsive Administration Theme

pandaskii:
Adminimal is the most popular contribute Administration Theme for Drupal

Based on the original seven theme, adminimal has several improvements like:
- Modernminimalist design that makes administration an easy and fun experience
- Responsive layout
- Usability Improvements
- Per site custom CSS code overrides without the need of hacking or subtheming the main theme
- More block positions and up to 3 column main content. So you can have left and right sidebars - blocks inside your content page
- Colored buttons for common tasks like Submit (Green button), Preview (Blue button), Delete (Red button) etc... So your eye rapidly focuses on the button you want to click
- Table highlighting. When you hover your mouse on a table row, it will be highlighted to help you see what goes where
- Improved navigation using the Adminimal Menu submodule
- And many more ...

See https://www.drupal.org/project/adminimal_theme

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -8,6 +8,8 @@ defaults[projects][subdir] = "contrib"
 projects[accessible_forms][version] = "1.0-alpha1"
 projects[acquia_connector][version] = "2.15"
 projects[addressfield][version] = "1.1"
+projects[admin_menu][version] = "3.0-rc5"
+projects[adminimal_admin_menu][version] = "1.7"
 projects[admin_views][version] = "1.6"
 projects[agls][version] = "1.0-beta3"
 projects[bean][version] = "1.9"
@@ -183,6 +185,7 @@ projects[xmlsitemap][version] = "2.3"
 ; Themes
 projects[zen][version] = "5.5"
 projects[shiny][version] = "1.7"
+projects[adminimal_theme][version] = "1.24"
 
 ; Libraries
 


fiasco:
I would only agree to adding another admin theme if we had an exit strategy for the incumbent: Shiny. So this PR should also include and update hook that:
- Disables the Shiny theme
- Enables Adminimal
- Sets Adminimal as the default admin theme **if** it were previously Shiny. We should not break user's choice to be using other admin themes they've added to the code base.

We'd also need to document an upgrade path to follow and include it in the release notes.

nathan-w:
If we're going to change the default admin theme for SaaS we need to do some Comms around this and show people what it will look like before it's rolled out.

seanhamlin:
I have used Adminimal before (years ago now), and it was real nice. I totally agree, if we are going to add this to the distribution, it should be with the view of deprecating (and ultimately removing) shiny.

Testing will be needed on more complex admin screens, e.g. node edits with heavy use of paragraphs etc.

Also, the installation profile should default the admin theme to Adminimal.

mgdhs:
This is a great idea. It is something we're using on our PaaS platform. Going back to SaaS without it feels so dated.

Why not do a staged release? Put it in the repo; let people try, use and test it; announce it; then do the forced switch; and finally remove the old theme.

That way you bring people along on the journey, it isn't a shock and this PR isn't wasted.

seanhamlin:
Hey @mgdhs I think that is a great idea, have phase 1 which includes the theme, and lets agencies adopt it if they want, and then phase 2 which mandates it (provided it goes all okay testing wise).

fiasco:
Agreed with @mgdhs suggestion. This PR just needs to set the new admin theme as default on install and we'll merge it in.

paulkilla:
Might be worthwhile also checking what sites have admin themes that use shiny as the base theme.

I know that we have a couple that use it as the base theme to get around annoying bugs with font awesome, and the WYSIWYG.

xtfer:
Is it necessary to remove the old one? Just add the new one and leave the old one alone.

paulkilla:
I also think we should think about adding admin_menu which integrates nicely with adminimal.
(and possibly setting this as the default setup for admin theme, (adminimal + admin_menu - navbar)

xtfer:
By way of context, the original reason for Navbar was that it was responsive, while the admin menu was not.

paulkilla:
Sounds like a reasonably decision was made.

I've just had a look using adminimal and admin_menu, with the addition of adminimal_admin_menu module (which is documented as a recommended addition to the theme) extends and makes it responsive!

paulkilla:
and @jozhao has already added that adminimal menu module as part of this PR!!

barbun:
I don't think the removal of Shiny is a good option here as some govCMS sites use it as the base theme for admin. So perhaps just disabling it would be the easiest.

paulkilla:
@barbun I agree, also the fact that content authors are also comfortable with that.. and at times they aren't always the most welcoming people to accepting change. :-)

fiasco:
If Shiny won't be exiting the project, then I'd advocate against adding another admin theme. Its creating more surface area to support for very little gain/value.

DJMyles:
We have (DHS) used Adminimal for around 9 months now and we've found it responsive and accessible and superior to shiny in terms of meeting our needs. @fiasco I agree with the approach outlined here in that we make it an option (include the theme) but don't force it to disable shiny in lieu of adminimal for phase 1. If it's widely adopted then phase 2 could look at setting it as the default admin theme.

alex-page:
As an enhancement of this ticket I would recommend bundling Adminimal theme with the [Adminimal Administration Menu](https://www.drupal.org/project/adminimal_admin_menu). This adds a consistent experience across the theme and the menu with advanced responsive functionality.

pandaskii:
Good idea https://github.com/govCMS/govCMS/pull/367#issuecomment-254695071

pandaskii:
@tobybellwood May I suggest this PR in the new release?

fiasco:
There are several reasons why I don't think we should pursue this PR:

1. If people want this theme, agencies can add it to their theme repo in the SaaS platform
2. This is only adding additional support responsibility for marginal gain - especially so late in the D7 software life cycle.
3. Admin menu conflicts with navbar which is already established and used. From a support perspective its more advantageous for everyone to have a common admin experience and Shiny wouldn't be going away any time soon.

PR #376, opened 2016-11-03
DRUPAL-SA-CONTRIB-2016-060: Security update to Workbench Moderation

seanhamlin:
…-060. This new major version includes a new dependency on a module called drafty, to which will be enabled by an update hook.


paulkilla (dismissed):
This look fine to me.

Can see the hook here to enable drafty:
http://cgit.drupalcode.org/workbench_moderation/tree/workbench_moderation.install?h=7.x-3.x#n677

We should also look into this: https://www.drupal.org/node/2579205

The 3 current issues relating to workbench moderation 7.x-3.x (https://www.drupal.org/project/issues/workbench_moderation?text=&status=All&priorities=All&categories=All&version=7.x-3.0&component=All) all relate to additional revisions.

seanhamlin:
Yeah, it seems like the excess revisions is on purpose, and the patch in https://www.drupal.org/node/2579205 might be needed to keep this under control (cleans them up on cron).

Chris820:
We updated this module - and taking on the new Drafty dependency - on our PaaS sites (in dev only at this stage) yesterday. We're also experiencing troubles.

In particular, with the scheduled publishing of nodes:
- Scheduler module: Still seems to work ok.
- Workbench_scheduler module: Does not appear to work
- Scheduler_workbench: Not yet tested.

Edit: Seems that each new revision to a node creates both a new revision for that draft, and another new revision (albeit unchanged) for the previously published version. :/
So i figure it'll have an impact on any module that relys on the $node->vid.

Also. A bit peeved off that a security update includes a new dependency on a module that is at most in a beta release. Good onya Drupal.

aleayr:
> Also. A bit peeved off that a security update includes a new dependency on a module that is at most in a beta release. Good onya Drupal.

Agreed @Chris820, it's annoying, feels lazy not releasing a minor point release to mitigate this single issue, and just bundling it with a major update that had some significant changes.

seanhamlin:
Linking this PR with this other PR as they are linked https://github.com/govCMS/govCMS/pull/379

aleayr:
@seanhamlin I'm not seeing the link between the Paragraphs issue (#379) and this one. Can you clarify?

Is the paragraphs draft issue caused by us not being on workbench_moderation version 3.0?

seanhamlin:
Hey Adam, well the link is that the paragraphs issue (in #379) is specific to workbench_moderation (7.x-1.4), and this thread is about updating workbench_moderation.

aleayr:
Ah, right. Good stuff. Cheers for the clarification.

seanhamlin:
Also noting #388 issue around Jenkins failing builds because of this PR.

pandaskii:
Note this issue at Drupal https://www.drupal.org/node/2826178

>Since updating Workbench Moderation to 7.x-3.0 due to a security issue, which also required the installation of a newly-dependent module "Drafty", Scheduler Workbench does not, well, schedule. There's a message that the content will be published at the designated time, but the content is published immediately. - https://www.drupal.org/node/2826178

Chris820:
I had a quick yarn to @aleayr about this at the Mega Meeting.

I'd suggest to have a look at the Workbench_Scheduler module. Specifically the 7.x-2.x-dev version. We're running it on our PaaS sites, and it works nicely, even with 3.0 of Workbench Moderation and Drafty.

I understand that because the Scheduler module is already within the distro, it doesn't make sense to have two scheduling modules. But the Scheduler module does suffer other issues (unpublishes nodes when revisions are scheduled) that the point of Scheduler_Workbench is meant to remedy.

Workbench_Scheduler, on the other hand, addresses both problems: Allows for scheduling of revisions, and plays nice with Drafty.

duartegarin:
Hi,
It seems that this issue has been open for quite a long time.
Can I understand the status on this? I am working on the PaaS version and the concept of releasing with security openings is unacceptable considering our client.
What is the proposed approach?
Thanks,

tobybellwood:
Hi @duartegarin - It's a fine mess, but basically the bump to 3.x introducing a whole new workflow, a dependency on Drafty (and breaking Workbench Scheduler integration amongst a number of other things documented here and in #392) have made this a herculean task to even consider it's adoption.  Given the volume of sites using our distribution relying on these workflow steps and schedules we have to take such potentially breaking changes seriously.  Whilst the damage to scheduling could be mitigated by a move to Scheduler Workbench, we would be unwilling at this stage to migrate to an "Unsupported" module.

The initial vulnerability is mitigated by the fact that any content lists must be regenerated at exactly the moment when a person saves the node, changing it's state.  We've taken the view that the chance of the vulnerability occurring, in conjunction with the low potential risks involved with exposing unpublished content (given the UNCLASSIFIED nature of our managed platform) the volume of work, re-work and upheaval to all of our agency clients to accommodate this version would not be the best expenditure of our resources.

Ideally we (and a seemingly large chunk of the Drupal community) had been hoping for a non-breaking patch on the 1.x branch, but given the passage of time, this seems unlikely.

Happy to discuss the potential implications of ignoring the vulnerability for non-managed (non-SaaS sites) if you want.

Chris820:
We run GovCMS PaaS, but not the GovCMS distro. Rather a vanilla Drupal install with whatever modules suit the requirements for the website.

We're using the Workbench Moderation, Drafty, & Workbench Scheduler combo.  It's great. Works well enough, and no warnings about out-of-date modules. 

But I understand that as the distro already includes the Scheduler module, a shift to a different scheduling module would be disruptive. So am happy to leave the decision to the GovCMS team.

Feng-Shui:
If anyone's considering moving ahead of the profile, we've recently upgraded workbench to version 3 with Drafty and have had no issues.  Obviously our configuration may be different to others (and we're PaaS) but thought it worth mentioning.

tobybellwood:
This is on our roadmap as something to think about - along with some other "all hands UAT" type activities (i.e. ones that agencies will have to assess their own impact for - the nuances will mostly be out of the public eye and hard to pick up, even with Visual Regression Testing)

PR #381, opened 2016-11-15
Updated Panels and Ctools modules to fix bugs.

barbun:
As a follow up to #331 after the security updates the Panels 3.7 module still has broken in place editing functionality for Site Builders.

Original issue reported here: 
https://www.drupal.org/node/2790499

Panels module has got a new stable release 3.8 that resolves this and many other bugs.

Note: The update also includes Ctools 1.11 upgrade.


seanhamlin (approved):
PR is simple, and having ctools bump with panels just makes sense. Keen to get this in.

barbun:
Ok, so this appears to be a bit bigger than I thought.

In current govCMS seems like no one except user 1 (even those who have admin role) can modify panels content (except for page and term tasks).

screen shot 2016-11-17 at 15 04 47

fiasco:
just noting that build failed because of werkbench module security update.

aleayr:
Thanks for reviewing Sean, merging.

Venkata-modgu:
Test passed.

Issue #388, opened 2016-11-24
Workbench Moderation module leads to Jenkins failure


pandaskii:
Recently PRs are facing a Jenkins failure message as following:

![snip20161124_3](https://cloud.githubusercontent.com/assets/1326673/20584818/55fbe1ec-b24a-11e6-8dfb-a7fed051b982.png)

As suggested from Workbench Moderation module

> The 7.x-3.x branch is the recommended version and the successor of the 7.x-1.x branch. If you're using a 7.x-1.x version you should consider updating to the 7.x-3.x as it fixes following security issue: SA-CONTRIB-2016-060

tobybellwood:
Merged in #492

PR #389, opened 2016-11-24
#388 Workbench Moderation module update

pandaskii:
REF #388


seanhamlin:
See another earlier PR for more discussion on this. https://github.com/govCMS/govCMS/pull/376 More testers are needed to ensure we understand any issues on what will impact likely all sites on govCMS.

PR #391, opened 2016-11-30
Rich text editor - valid elements update

aleayr:
This PR fixes a discovered bug where new sites don't have access to use video, audio, source, address or abbr tags.


fiasco (approved):
Looks fine to me as its just addding in the use of HTML5 elements in the WYSIWYG editor. It also allows all attributes which I'm unclear on if that poses any risks. If it does, I suggest we address that in later PRs.

Venkata-modgu:
Fix issue #152
Fix issue #153

PR #397, opened 2016-12-20
Enforce the default list of file extensions when uploading files and creating content types

invisigoth:
As a platform administrator, I would like the default list of file extensions to be enforced when uploading files and creating content types


ruwanl (approved):

Venkata-modgu:
Test passed.

PR #404, opened 2017-01-09
[GOVCMS-UI-KIT] Updated bundled GovCMS Features to work with GovCMS U…

ivangrynenko:
This is alteration of GovCMS bundled features to apply the changes described in https://salsadigital.atlassian.net/wiki/display/GOVCMSTHM/govCMS+profile+changes

Note, that lines 15 and 18 from the Confluence page https://salsadigital.atlassian.net/wiki/display/GOVCMSTHM/govCMS+profile+changes need to be applied manually, since they are block positioning and will depend on the final design.

This PR does not include the theme itself.


deanvucic:
This will not be included in the upcoming milestone. Closing PR.

Issue #405, opened 2017-01-10
Add ability to use 3rd party SMTP service


aleayr:
govCMS should support accessing an external SMTP service, as only basic transactional emails should be sent from the Acquia platform (and in general, it isn't best practice to send email from a web server).

aleayr:
Interested in feedback before I go ahead and merge this in.

aleayr:
- What are the ISM implications?
- Should govCMS enforce TLS to SMTP servers?

Podgkin:
Preferably supporting TLS, DKIM and SPF as per ISM.   Thing is though, any agency that turns this module on will have to be responsible themselves for meeting ISM on their implementation.  govCMS can have sane defaults; but the moment someone sets up their own SMTP implementation, it leaves the scope of a software product and becomes an infrastructure configuration issue.

Feng-Shui:
Agree with @Podgkin that this is really a infrastructure issue.

TLS is a no brainer if you're passing from cloud to a locally hosted SMTP server, but if you're running Drupal in a DMZ you might not pass things around internally using TLS.  A sane default config is hard to figure out.

I don't have any concerns with this functionality going in, I think it provide necessary configuration flexibility - esp for those people who self host.  And even on cloud, routing things somewhere else is useful (esp for testing).

aleayr:
Some good points, thanks @Podgkin and @Feng-Shui.

At this stage, I'm going to hold off putting this in, pending a stronger use case.

We'll revisit it if a few customers require this. Self hosted customers always have the option of adding the module specifically if they require it.

PR #406, opened 2017-01-10
Added SMTP module

aleayr:
Allow customers to use an external SMTP service if they require to send anything other than transnational email from Drupal.


aleayr:
#405

aleayr:
Interested in feedback, this is the recommended module from Acquia though:
https://docs.acquia.com/cloud/configure/email/3rdparty

Issue #407, opened 2017-01-12
Module Update: User protect


cjoneill:
Acquia received a support ticket from an agency using govCMS on the Site Factory platform (govCMS SaaS).  Ticket reference for those with access: 337760

The deletion of content triggered notices/warnings in `userprotect_get_user_protection()` as seen in the following screenshot:

![Screenshot of notices from userprotect module](https://p5.zdusercontent.com/attachment/18825/uIu18bx88U8nkyZQZwb9yiPGl?token=eyJhbGciOiJkaXIiLCJlbmMiOiJBMTI4Q0JDLUhTMjU2In0..qM_0n8VnZ_7Eo8xlXAVERw.FIT9fPqRjBczpZWetL6ce6LCG6W5yRu2_kzK0tHwlUmuWFKx4Mn__k9QCTBG7C5XJEPXyV0zbI3lS9XPoltWb8WMeArOxi3mMlniWH9oxvxzDYx-CngLXd_mRNJdLktC_yXI6XOTzudGD4z_rRDQFmtsYxUVRAE0pBo3yrCf2DC81LT6L3EcRYjSqszAZmsB6hUWvWcFTYXHzWZDOYx4n96mrhF13_0SkdlPADFE69u2l8w5VjvL_r4E9qUoN7tmGdGCf7jnb5Cp9DUupFMCxB94DklxP5xCHzygmd_1NZU.6wtf19njDECMhurwnPLF1Q)

This may be caused by a since-fixed bug in userprotect-7.x-1.1 (release August 2014). 

Patch was committed to contrib in Nov 2014: https://www.drupal.org/node/2380591

 7.x-1.2 was released in August 2015: https://www.drupal.org/project/userprotect/releases/7.x-1.2

Could this module be considered for updating in the base distro?

aleayr:
Thanks @cjoneill - definitely worth looking into.

aleayr:
PR: #494

seanhamlin:
PR is merged.

tobybellwood:
Closed as PR merged

PR #408, opened 2017-01-16
User protect module update

aleayr:
As per #407, this update solves a bug where warnings will be shown to the user.


seanhamlin:
Just reviewed the code delta between 1.1 and 1.2, and it is very light, and only bug fixes. http://cgit.drupalcode.org/userprotect/log/

Plus https://github.com/govCMS/govCMS/issues/407 seems pretty bad. Seeing as Adam raised, this, I am happy to merge it.

seanhamlin (approved):
Just reviewed the code delta between 1.1 and 1.2, and it is very light, and only bug fixes. http://cgit.drupalcode.org/userprotect/log/

Plus #407 seems pretty bad. Seeing as Adam raised, this, I am happy to merge it.

Issue #458, opened 2017-01-23
Add more services to govCMS Social Links


nathan-w:
As a content publisher 
I want to add my official LinkedIn, and Instagram accounts to the Social Links block
So my users can easily find the accounts and the great content we have

-----------
We currently support Facebook, Twitter, Vimeo, Youtube etc - LinkedIn and Instagram are common enough we should be able to add those too

DJMyles:
Great chat re: social media block implementation. Good opportunity in the future to make social media platforms more fluid in terms of actual platforms, their attributes and how they might be output by govCMS adopters (blocks, views, cards etc).

I like the taxonomy approach where you have a social media vocab with a bunch of platforms as terms, each containing fields such as name, URL, description, an icon etc. Provides limitless flexibility in terms of how to manipulate and control the output of that taxonomy (block, bean, card, list etc). You then get the added benefit of creating powerful relationships to other content by utilising entity references.

seanhamlin:
I believe this is addressed in the PR https://github.com/govCMS/govCMS/pull/460, so this issue can be closed.

PR #459, opened 2017-01-23
#458: Add more services to govCMS Social Links

aleayr:
Instagram and LinkedIn added to govCMS Social Links module.


nathan-w:
Do we need to include default icons too? Or is that dealt with on a theme-by-theme basis?

Issue #378, opened 2016-11-04
Consultation feature does not accept certain valid Australian phone numbers


invisigoth:
It has been noted the "phone" module on which the consultation feature is based does not allow valid Australian phone numbers such as 0261xxxxxx (TAS) and 025xxxxxxx (ACT) and some mobile numbers.

Wongad:
This is the same for the Phone field in the Webform as well, not just the Consultation feature.
We've workaround it by using just a Number field, with integer, not thousand separator and uncheck the Exclude zero

Feng-Shui:
PR up to provide support for the missing numbers #463

PR #463, opened 2017-01-30
Add patches to Phone module to support newer Australian numbers

Feng-Shui:
Issue #378

There are RTBC patches up on D.O that provide support for additional Australian numbers:

https://www.drupal.org/node/2723821
https://www.drupal.org/node/2050917

This PR adds these two patches to the make file.


deanvucic:
This will not be added in the next release. Closing PR.

Issue #263, opened 2016-06-10
TinyNav select doesn't have a label


alex-page:
`<select id="tinynav1" class="tinynav tinynav1"></select>`

This is an accessibility concern.

An example of a patch to fix the issue can be found here:
https://www.drupal.org/node/2204833

A temporary fix for this can be by adding some javascript to the theme:

```
$( document ).ready(function() {
  $('#tinynav1').before("<label for='tinynav1' class='element-invisible'>Mobile menu</label>");
});
```

aleayr:
We should encourage this patch to be accepted upstream I imagine? Been there for 2 years though, which is a worry.

fiasco:
tinynav is perhaps a good example of a bad module to adopt for a distribution as it doesn't have the upstream support we need.

But since we're stuck with it, we can choose to adopt the patch.

fiasco:
If we can get a PR to apply the patch we'll review to dd to govCMS. Thanks!

alex-page:
@fiasco I have created a PR for this #467

PR #467, opened 2017-01-30
Accessible TinyNav - Includes relevant label above <select> element

alex-page:
Making tinyNav accessible by adding a relevant <label> tag for the <select> dropdown.

#263


deanvucic:
This will be considered for future milestones. Closing PR.

Issue #461, opened 2017-01-25
Allow more output options from views


aleayr:
# Full Description
Currently, views supports RSS output. A number of agencies have asked for further options such as XML, JSON/JSONP options.

# Acceptance Criteria (Must)
- I must be able to generate XML, JSON/JSONP output from the views module; and
- I must be able to consume the resultant output from a separate govCMS website, service or application.

# Acceptance Criteria (Desirable)
- I would like to be able to modify/map the output of XML and JSON in particular to match a particular format required by different external applications.

# Proposed Solution
[Views Datasource](https://www.drupal.org/project/views_datasource) allows an extensive amount of output covering the requirements listed above, and additional output formats including ATOM and other microformats.

Research as to the desirable acceptance criteria, or alternate options, covering all use cases, appreciated.

aleayr:
Worth looking into https://www.drupal.org/node/633702 and http://drupal.stackexchange.com/questions/22878/preprocess-views-view-specify-view-name/53388#53388 in relation to whether this would allow modification of XML/JSON to a specific format.

Podgkin:
Those acceptance criteria sound an awful lot like both Content Hub and the Services module

aleayr:
True, this serves a different use case though. Content Hub isn't a 'thing' at this stage, and Services module is more of a developer option. This solves a number of issues Site Builder customers have been asking us about.

PR #470, opened 2017-02-13
Added views_datasource module

aleayr:
#461


aleayr:
Testing has passed.

govhosting (approved):

aleayr:
Whoops, wrong issue commented on. I'll revert this, should be in the subsequent release.

aleayr:
Given this is a very minor change to the output options for Views (allows JSON and XML in read only format) in addition to the RSS that is already there, I might just get Venkata to test this in the test environment so I don't have to do a revert at this stage.

Venkata-modgu:
Test passed.

PR #472, opened 2017-02-21
Update govCMS CKAN to 1.2

steveworley:
Minor bug-fixes and performance optimsations for govCMS CKAN. With the launch of DoE the platform should be running the latest version of the module.

- Fixes IE11 downloading issues
- Adds query limit to prevent OOM issues when requesting large datasets
- Adds per file cache settings for each visualisation
- Adds global default cache setting
- Updates the max length of the URL field


seanhamlin:
Full changelog https://github.com/govCMS/govcms-ckan/compare/7.x-1.1...7.x-1.2

aleayr:
Sean, as discussed, if you could assist with testing, that would be great.

aleayr (approved):
Approved for release to environment for integration testing.

PR #482, opened 2017-03-20
Update ctools, panels and views to the latest stable release, also pa…

seanhamlin:
…tch ctools to avoid cache bins filling up as per https://www.drupal.org/node/2828620.


aleayr (approved):
Internal testing has shown this can be merged into the codebase for full integration testing.

PR #483, opened 2017-03-20
Add default media permissions for editors

barbun:
I believe this is an issue that is being bumped into with every govCMS project. Time to fix it once and for all!

Out-of-the-box govCMS does not provide even Site Editors with the ability to use Media Browser and Media WYSIWYG. 

To replicate:
For example, as a Site Editor try to add an Image or File when creating / editing a Standard Page node. You will be presented with a big fat "You are not authorized to access this page.". 
Likewise, if you try and add an image via WYSIWYG you will face the same issue. , which lies in "Use the media browser" and "Use Media WYSIWYG in an editor" permissions.

The same holds true for all of the roles below Site Editor, which is really something I personally would not consider as an expected behaviour as adding Images and Files should be a standard content authoring process.

Moreover, after giving the users access to use Media Browser, I figured that Content editor does not have permissions to "Add and upload new files", which is again something that I would consider as an expected default. (Note, Content approver doesn't have this permission either but that seems sound since the sole role of approve is not to create content but approve).

Moving forward, after granting permission to "Add and upload new files"  both Content editor and Content approver do not have permissions to "Edit own files", "Delete own files" etc. which again creates obstacles for providing additional metadata around the File Entity (e.g. Alt field of an image).

To address all these issues, I've updated OOTB govCMS Editing feature with a set of permissions that in my view should be default. Even though Site Builder and Site Editor have "Bypass file access control" permission I thought it wouldn't hurt to provide standards for them as well.

Happy to hear out any objections. Otherwise, this would simplify life for a lot of people out there ;-)


aleayr:
I'm always a fan of more default permissions updates.

@govCMS/osb-devs Are there any other permissions updates you're aware of we could bundle with this, asides from the work @paulkilla did on #395

gavintapp:
Adding my vote for this PR. Would be a nice timesaver when setting up new sites.

barbun:
Test this

barbun:
Test this

barbun:
Test this

barbun:
This passes the tests, however, due to recent updates in Jenkins the "default" state that is marked as failed is no longer valid and won't be used for future PRs.

Not sure if "default" check can be removed manually for this (and other open) PR?

modules/features/govcms_editing/govcms_editing.features.user_permission.inc
```diff
@@ -0,0 +1,334 @@
...
+  // Exported permission: 'delete own image files'.
+  $permissions['delete own image files'] = array(
+    'name' => 'delete own image files',
+    'roles' => array(
+      'Site editor' => 'Site editor',
+      'administrator' => 'administrator',
+    ),
+    'module' => 'file_entity',
+  );
+
```
ruwanl, line 111: Is there a reason why this permission is only set to site editor and admin instead of all four roles like above?

modules/features/govcms_editing/govcms_editing.features.user_permission.inc
```diff
@@ -0,0 +1,334 @@
...
+  // Exported permission: 'delete own video files'.
+  $permissions['delete own video files'] = array(
+    'name' => 'delete own video files',
+    'roles' => array(
+      'Site editor' => 'Site editor',
+      'administrator' => 'administrator',
+    ),
+    'module' => 'file_entity',
+  );
+
```
ruwanl, line 121: and this one?

modules/features/govcms_editing/govcms_editing.features.user_permission.inc
barbun: No particular reason, this was a "minimal" approach as I wasn't sure what are the rules around deleting the files in govCMS. It might make sense not to allow such operation for traceability. Another example would be, if Content editor adds an image to a node and the node subsequently gets published then that same editor will have the permission to delete that image from the system - something that could be considered risky.

ruwanl: Thanks @barbun.  To be consistent should we have the same applied to "delete own document files" and "delete own audio files" or is your thinking different in these cases?

barbun: You are right @ruwanl , I missed this consistency in my PR. Thanks! 
I can adjust and re-submit this later today.

deanvucic:
This will be considered for future milestones. Closing PR.

PR #522, opened 2017-07-11
Automated testing

barbun:
Behold TL;DR! This is a rather big chunk of work that contains significant enhancements to the existing govCMS automated build and testing tools. The majority of the changes is focused on Behat testing with some other updates tagging along to enable for broader experience with the test scenarios.

**Key changes:**
- Increased testing for the govCMS distribution, which will provide greater confidence when merging PRs. 
- Expanded on the ways in which tests can be written
- Provided a framework that can be extended to write more tests
- Updated the toolset to stable versions
- Fixed (or indicated) some general issues found along the way (see “Additional changes”)

The test coverage is now estimated at approx. 75% of the functionality with the number of Behat tests increasing from 16 to 38 with addition of 1 PHPUnit test.

# Build Tools
The govCMS build toolset has been upgraded with newer versions to benefit from the latest features, solutions and bugfixes. The *composer.json* file enclosed in build folder is responsible for assembling all the needed parts to run the build process. The following changes were implemented to composer.json:
- Updated Drupal Behat extension to latest stable release
- Removed Drupal Driver library as it became a dependency of Drupal Behat extension
- Updated Goutte screen scraping library for PHP to latest stable release
- Removed Guzzle PHP library because it is now a dependency for Goutte
- Removed PHP CodeSniffer library as it is now pulled as a dependency
- Updated PHPMD library to latest stable release
- Removed PDepend library as it is now pulled in as a dependency of PHP Depend
- Updated PHPLOC tool to a stable version. Note, later versions exist but need testing.
- Updated PHP Copy/Paste detector to latest stable
- Added PHPunit 5.7.

As this combination of tools and libraries can be fragile, for future maintenance and updates the newer versions should be tested to work together and provide the expected results before committing them into *composer.lock*

The composer.lock file has been updated accordingly to lock the tested versions of the tools.

# Build script
GovCMS uses Phing as a PHP project build tool. All the custom steps and logic for creating govCMS project are scripted in XML format and contained within *build/phing/build.xml* file. The script has been augmented to add two additional steps: 

- Run Behat tests with two different profiles (explained below)
- Run PHPUnit tests (explained below)

# Drupal API driver
Previously, Behat was running using Drush driver (read more [here](http://behat-drupal-extension.readthedocs.io/en/3.1/drush.html)). As the scenarios advanced it became necessary to introduce Drupal API driver (read more read more [here](http://behat-drupal-extension.readthedocs.io/en/3.1/drupalapi.html)) to unlock the full potential of testing. In a nutshell, the Drupal API driver allows Behat to create users, roles, nodes, vocabularies and taxonomy terms on the fly. The only drawback of such, compared to others, is that it cannot be run on a different server than the site being tested. This is not an issue with the current approach for govCMS builds using Jenkins.

This resulted in re-writing the default Behat profile (*build/tests/behat/behat.yml*) to use Drupal API driver.

# PHPUnit support
Support for PHP unit tests has been added, with the *phpunit.conf* and *bootstrap_phpunit.inc* files being placed in build/tests/phpunit. Tests are provided with a fully bootstrapped Drupal environment and will use the default site's database (installation should have run).

The tests make use of the TDD7 (Test Driven Development for Drupal 7), a module hosted at https://github.com/edward-murrell/tdd7, which provides mock versions of a good number of Drupal functions. TDD7 doesn't need to be enabled in the Drupal installation. Some additional mocks were written and these could be submitted back upstream (although the project seems to be inactive).
With the added configuration, docroot is not checked for tests. Rather, the modules directory next to the docroot and its contents are searched, so tests will only be run for the custom govCMS modules. Files containing PHPunit tests should match the pattern '*Test.php'.

As the pilot, Unit tests have been written for the Iconomist module and together with TDD7's self tests they can be invoked from the parent directory of the build with the following command:
```
build/bin/phing -f build/phing/build.xml test:phpunit
```
The iconomistTest.php may be used as a pattern for writing new tests for other modules - it shows how to include TDD7, mock Drupal core functions and implement tests. The heart of the implementation is the use of a namespace, within which functions are defined with the same name as those being overridden. Since the invocations in Drupal itself aren't prefixed with namespaces, the definitions in the file's namespace can override them. These then invoke the static methods defined in TDD7.

The *phpunit.conf* file includes configuration for generating code coverage information as well. This is commented out by default as it's not needed for Jenkins. When enabled, the output will be placed in a */coverage* directory (next to docroot).

# Custom Behat steps
The Behat/Mink and Drupal Extension provide the standard set of step definitions that interpret the human-readable language into a logical code to perform the task. For example:
```
Given I am logged in as a user with “Content editor” role
```
This step is defined by a function within the Drupal extension; it takes “Content editor” as an argument, creates a basic user with that role assigned and logs that user in, so the scenarios can be performed from a perspective of Content editor, in this particular case.

Behat supports custom steps definitions via so called Contexts (read more here). The custom steps can be used to further expand on the existing contributed steps or add brand new definitions that can be tailored to specifically fit your system. A number of custom step definitions have been added to provide broader set of typically performed activities and cover some of the govCMS specifics. 

GovCMS already has a few custom step definitions that can be found in *build/tests/behat/bootstrap/FeatureContext.php* file. Some of those steps are slightly modified copies of the defaults from MinkContext or DrupalContext.

Due to a relatively big number of custom steps, it made a lot of sense to sort them out by some criteria. This was achieved with the sub-contexts approach available in Behat. The following sub-contexts and the steps that they supply (examples) were added to *build/tests/behat/bootstrap*:

1. **AccountContext.behat.inc**
```
Given I am logged in as a user named “John” with “Content editor” role
```
```
Given I am logged in as a user named “John” with the “administer site content,administer site configuration” permissions
```
```
Given I am logged in as a user with the password “ABC123” and the “administer site content” permission
```
```
Given I am logged in as a user named “John” with the following permissions: (list)
```
```
Given a user named “John” is deleted
```
```
Given a user named “John” has not logged in for “58” days
```
2. **ChosenContext.behat.inc**
```
When I set the chosen element “Tags” to “mytag”
```
3. **CkEditorContext.behat.inc**
```
Given CKEditor/WYSIWYG for the field “Body” exists
```
```
Then CKEditor/WYSIWYG for the field “Body” should exist
```
```
And I put “Hello” into CKEditor/WYSIWYG (of “Body” field)
```
```
Then CKEditor should contain “Hello”
```
```
Then CKeditor should match “/^def/” (regular expression)
```
```
When I execute the “bold” command in CKEditor
```
4. **MediaBrowserContext.behat.inc**
```
When I open the “Image” (field) media browser
```
```
When I submit the media browser
```
```
When I open the media browser for “Attach media” in “Documents” field set
```
```
When I select item “1” from the (id) media browser
```
5. **PermissionsContext.behat.inc**
```
Then the “Content editor” role should have permission(s) to:
“””
Access media browser
Administer site content
“””
```
```
Then the “Content editor” role should not have permission(s) to: (list)
```
```
Then the “Content editor” role should have the “access site configuration” permission
```
```
Then the “Content editor” role should not have the “access site configuration” permission
```
6. **MiscContext.behat.inc**
```
Then the checkbox named “enabled” in a table row with text “Activate the chaos” should be checked
```
```
Then the checkbox named “enabled” in a table row with text “Activate the chaos” should not be checked
```
```
Then I should see (the text) “Running” in a table row containing (the text) “Chaos status”
```
7. **UtilityContext.behat.inc**
```
Then I should not see a “Body” field
```
```
Then “3” elements should match “selector.list-item”
```
```
Then at least “3” elements should match “selector.list-item”
```
```
Then nothing should match “selector.list-item”
```
```
Then no elements should match “selector.list-item”
```
```
The the “Body” field should be empty
```
```
When I clear “Body” (field)
```
```
Then the “Category” select field should be set to “New”
```
```
Then the “Category” select field should not be set to “New”
```
```
Then “active” radio option should be selected
```
```
When I accept the alert(s)
```
```
When I select the radio button (“active”) with the partial id “govcms-option”
```
8. **FeatureContext.behat.inc**
```
When I schedule the node to be published at “YYYY-MM-DD HH:MM:SS”
```
```
Then I take a screenshot named “capture”
```
The contexts above also contain some of the useful helper functions that can be benefited from in the future. For instance, *FeatureContext.behat.inc* provides functions to resize the browser window, take a screenshot at any point etc.

As an enhancement to the existing Behat Drupal Extension, additional cleanup functions were written to remove any test files and blocks that were created during the scenario. 

Also, govCMS has a rule to force password change on first user login which would fail most of the standard tests. To bypass this rule, the login steps were previously customised to drop the requirement for forced password change, e.g.
```
Given I am logged in as a user named "peter" with the "Content editor" role that doesn't force password change
```
This was very inconvenient from the Behat perspective, especially when more steps involving the login operation arrived. It was decided to assume by default that the users do not need force password change, unless specified otherwise. This allowed for the steps to be more standardised and reusable, so it’s not stating anything govCMS specific:
```
Given I am logged in as a user named "peter" with the "Content editor" role
```
Additional backend logic was included in *FeatureContext.behat.inc* to handle the negation of force password change rule upon user login during the tests.
Some other changes and improvements to the existing custom steps have been included as well.

# Behat Scenarios
Note, the purpose of the added tests is to check that the govCMS distribution works as expected out-of-the-box. They aim to ensure that any change to the default govCMS configuration is brought to the attention, thus minimising the risks of breaking the existing functionality by introducing/altering the code.

As the number of scenarios made a significant increase, it made sense to sort them out into different folders. The proposed structure so far: Content types, Blocks, Configuration and Workflow. In the similar fashion, when more Behat scenarios will be produced in the future they can be either put into an existing folder or be organised into a new folder (e.g. Views, Panels etc.).

The following is a quick overview of new tests.

## Content types
All of the default govCMS content types have been covered with the basic sets of scenarios. These usually perform such steps as: Verify that WYSIWYG is available; Create sample content using the available fields (including attaching of media and files); Check that the page is created and displays expected information (including image styles and attributes); Check that the content also looks intact in other places where it appears (if applicable, e.g list of latest news); Confirm that breadcrumbs work (if applicable); Ensure that the Moderation workflow is available and works (where applicable); Other misc configurations (eg. custom menu link creation, promotion to front page etc.) 

The following content types scenarios were added:

- Blog Article
- Event
- Media Release
- News Article
- Promotion
- Publication
- Standard page (enhanced the existing)
- Slide

## Blocks
Similar to the content types, the scenarios were written for out-of-the-box blocks:
- Basic bean
- Image and Text bean
- Menu block
- Social links

The tests assert each block type to the expected behaviour and configuration so to ensure they are intact and properly configured.

## Configuration
These tests are focused on custom govCMS configurations as they are set during the installation and via custom modules/features that ship with the distribution.

- Account suspension
- Canberra timezone
- Iconomist
- Password policy
- Permissions
- Share Links
- Theme settings
- User actions
- Contact
- Metatags (enhanced)

## Workflow
The workflow tests contain common user activities on the site and assert their expected outcomes.
- Reset password
- Text resize

# Additional changes
Various adjustments and additions have been made:

- Adjusted the existing Behat context code to conform to PHP and Drupal Coding Standards
- Added an image storage folder for Publication content type (*/images/publication*), so it follows the same principle as other content types
- Added TDD7 module to assist with PHPUnit testing.
- Added sample files to *build/tests/behat/files* that are being used by Behat to test file upload functionality
- Converted the Iconomist module to basic OOP principle to enable PHPUnit testing
- Updated contrib *password_policy* module from 1.11 to 1.14 to avoid the buggy behaviour during the tests.
- Tweaked custom *govcms_password_policy* module in accordance with the updates of password_policy contrib module.
- Adjusted *CONTRIBUTING.MD* to fix the wrong instructions about submitting PRs. (against 7.x-2.x branch and not master)

Some of the issues found along the way (not fixed as part of this PR):
- No permissions configured for Media (see https://github.com/govCMS/govCMS/pull/483). This prevents the tests that use Media from being executed under any user other than administrator. *@ToDo* comments have been added in the scenarios to adjust once 483 is merged.
- Publication and Event content types don’t apply an image style on full display
- Except Publication, no other content types have File Upload Size limit set, which by default would allow to upload files up to 256MB. This has a potential to quickly escalate for the Event content type as it outputs the original image on full view.
- Promotion content type has the machine name “footer_teaser”, which can be misleading.
- Footer teaser view on the homepage links the title of an item to a referenced entity but the image of the same item links to the promotion node itself.
- Custom govcms_register module doesn’t seem to be used anymore. The endpoint included in that module does not respond.
- Consultations Behat test enables the govcms_consultation feature but does not clean up after the test has finished (should bring it back to disabled as by default).

Hopefully this will further solidify the integrity of govCMS distribution, streamline any future changes to the distribution and provide a good starting point for the agencies who decide to leverage this work for the automated testing of their own site instances.

Note, because of a broader test coverage, some new PRs would also need to include the changes to the Behat tests. The information about changing and adding new tests have also been added to *CONTRIBUTING.MD* file


tobybellwood:
Thanks @barbun - glad to see the tests passed their own tests :-)

CONTRIBUTING.md
```diff
@@ -82,6 +84,65 @@ composer install --prefer-dist
...
+  - @javascript can be added when the scenario requires javascript
+  - @drush can be added when the scenario needs to be run using Drush driver
+
+*Note, scenarios without @javascript would usually be much faster, however they
+won't produce a screenshot on failure.*
+
+There are plenty of existing steps available already that handle some typical
+activities on the site so you might also want to scan through the scenarios to
+get a better idea. Additionally, you can have a look inside the
+*/build/tests/behat/bootstrap* to find the contexts that define some of the
```
seanhamlin: I think you need to be careful about asterisks in markdown. Please either:

* use inline code blocks where appropriate
* escape the asterisk into HTML entity `&ast;`

I think for describing folders etc, you should use the first option `*/build/tests/behat/bootstrap*`

CONTRIBUTING.md
```diff
@@ -82,6 +84,65 @@ composer install --prefer-dist
...
+
+The iconomistTest.php may be used as a pattern for writing new tests for other
+modules - it shows how to include TDD7, mock Drupal core functions and implement
+tests. The heart of the implementation is the use of a namespace, within which
+functions are defined with the same name as those being overridden. Since the
+invocations in Drupal itself aren't prefixed with namespaces, the definitions
+in the file's namespace can override them. These then invoke the static methods
+defined in TDD7. The phpunit.conf file includes configuration for generating
+code coverage information as well. This is commented out by default as it's not
+needed for Jenkins. The output will be placed in a */coverage* directory.
```
seanhamlin: More asterisk fun.

build/composer.json
```diff
@@ -39,5 +35,8 @@
       "post-update-cmd": [
         "PhantomInstaller\\Installer::installPhantomJS"
       ]
+    },
+    "require": {
+        "phpunit/phpunit": "^5.7"
```
seanhamlin: this should be in require-dev.

build/tests/behat/bootstrap/ChosenContext.behat.inc
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Acquia\govCMSExtension\Context;
```
seanhamlin: Unsure about this namespace, can we make this generic to govCMS?

build/tests/behat/bootstrap/MediaBrowserContext.behat.inc
```diff
@@ -0,0 +1,132 @@
...
+    }
+    $link = $element->findLink('Browse');
+    if (empty($link)) {
+      throw new \Exception(sprintf('Failed to find Media Browser in the fieldset "%s" on the page %s', $fieldset, $session->getCurrentUrl()));
+    }
+    $link->click();
+    $this->minkContext->iWaitForAjaxToFinish();
+    $session->switchToIFrame('mediaBrowser');
+    // This might be vestigial.
+    sleep(10);
```
seanhamlin: Do we need this? Or how about waiting for the modal to be visible?

modules/custom/tdd7/LICENSE
```diff
@@ -0,0 +1,281 @@
+                    GNU GENERAL PUBLIC LICENSE
```
seanhamlin: No need for GPL license files inside modules in a GPL distribution.

modules/custom/tdd7/docs/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Wrapping Namespaced Code
+Drupal hooks are in the global namespace, and test driven development requires
+us to keep our production code out of the global namespace so that we can use
+mock classes and functions.
+To get around this problem, we call our module code from wrapper functions.
+These wrapper functions should be a single line of code that passes all
+arguments, and return the results without making any changes.
+
+	<?php
```
seanhamlin: Using triple ```php is preferred as github adds syntax highlighting.

modules/custom/tdd7/docs/examples/tests/ExampleDrupalNodeMockTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+/**
+ * @file
+ * @author Edward Murrell <edward@catalyst-au.net>
```
seanhamlin: Should this module be brought in by composer? Rather than being inserted into the repo in it's entirety?

seanhamlin (changes requested):

modules/custom/tdd7/docs/examples/tests/ExampleDrupalNodeMockTest.php
seanhamlin: This review comment may supersede the above 2.

barbun: We can, but I was a bit concerned that if repository goes away then the build process would be failing.

pandaskii (changes requested):
Needs a clean PR only for automated testing

tobybellwood:
Closing - PR in #651

PR #531, opened 2017-07-17
Update bean to 7.x-1.11 (from 7.x-1.9)

tobybellwood:
## https://www.drupal.org/project/bean/releases/7.x-1.11

Release notes
7.x-1.10 shipped with broken views support caused by commit 3ab5e4e from #2527008: Display cache breaks bean display. This was rolled back as a result of #2670862: Views display as Rendered Entity has stopped working for Beans when using bean dev release.

## https://www.drupal.org/project/bean/releases/7.x-1.10

Release notes
Known Issues:
#2670862 Views integration is broken.

Issues Resolved:
#2527008 by axel.rutz: Display cache breaks bean display.
#2465267 by brantwynn: Bean insists we use a tagged release of Entity API and that's simply not happening.
#2576711 by a.milkovsky: Missing isNewEntity() method caused problems with e.g. entity_extract_ids(), Deploy, etc.
#2623448 by helmo: Ensure uuid.install is loaded when overloading its schema.
#1995770 by kris84, Albert Volkman: Use the label as the entity label, not the title. This should resolve a number of issues with other modules.
#2572517 by JeroenT: Weights of submit buttons on the bean form results in the Entity Translation 'Remove translation' button appearing before the 'save' button.
#2743457 by dan.munn, max.parker: Bean labels being encoded on admin pages
#2714693 by fullerja: In watchdog logging, use type rather than typeName()
#2702495 by plopesc: Get rid of bean_load_delta() in Bean entity class for better compatibility and performance
#2459041 by jsacksick: Use bean_access as access callback for block/add/(type)
#2267391 by freality: Bean properties display as improper type in Views


ruwanl (approved):

Venkata-modgu:
Test passed.

PR #545, opened 2017-08-07
SA-CONTRIB-2017-062 Update services_views to 7.x-1.2 (from 7.x-1.1)

tobybellwood:
fixes services_views - Unsupported - SA-CONTRIB-2017-062

Posted by Drupal Security Team on 3 Aug 2017 at 00:18 AEST
Advisory ID: DRUPAL-SA-CONTRIB-2017-062
Project: services_views (third-party module)
Date: 2-Aug-2017

## https://www.drupal.org/project/services_views/releases/7.x-1.2

Release notes
SA-CONTRIB-2017-062: Arbitrary Views exposed over API

This release provides a mechanism that allows you to secure view displays that are arbitrarily available via the Views resource. You will now be able to whitelist/blacklist specific view displays from being displayed.

This release also includes changes that makes people aware of potential view displays that could be leaking information. There are warnings in several locations including next to the Views Resource on each api endpoint and the status report. There is another full report showing all view displays without access control set up.

The Problem
The reason this security issue is a problem is because by default, panel pane displays, block displays, and several other type of displays, are not accessible via a "route" by default and would not be accessible to the public unless someone put them on a page that was accessible to the public.

The Solution
The suggested work around for now will be to use the new Views Resource Settings page located at http://YOURSITE/admin/structure/services/list/ENDPOINT_NAME/view_resource. Whitelist or Blacklist the displays you wish to be available to the endpoint. You will need to do this for each endpoint. For a good idea of where to start removing displays available to the endpoint, look at the report at http://YOURSITE/admin/reports/insecure-view-displays. These settings are saved as simple drupal variables, so you can use Strongarm to export them if you need.

Important Next Steps
This is a single step in a bigger initiative to remove the 'views' endpoint all together and make people aware of the security implications of reusing view displays that typically do not have a non-administrative endpoint to them such as panel panes and blocks. By September 4th, 2017, the views resource will be removed unless there is significant cause to not do so. All white listed view displays will be copied and transformed into a "Services" type view display. the url pointing to the views resource will be backward compatible to these new displays only as long as the views resource is enabled. The goal is to move away from this all inclusive insecure resource to help developers keep information from being leaked to unintended users of the api.

The progress will be followed in #2900460: Remove the Views Resource.


ruwanl (approved):

Venkata-modgu:
Test passed.

PR #553, opened 2017-09-06
Update date to 7.x-2.10 (from 7.x-2.9)

tobybellwood:
## https://www.drupal.org/project/date/releases/7.x-2.10

Release notes
Fixed below bug/feature fixes:

#2377623: Remove unnecessary space in data_combo title
#2186191: missing help text
#2412049: DateMigrateExampleUnitTest fails if migrate is missing
#2646646: Tests are broken on current 7.x-2.x branch
#2780415: As per Drupal coding standards, use uppercase for PHP constants
#2648114: DateTime::format(): The DateTime object has not been correctly initialized by its constructor in DateObject->format() 
#1037150: date_popup element with just time doesn't work
#2160843: Grammar issue: "Date objects can be created and cached as date fields are loaded rather than when they are displayed to improve performance"
#2611968: PHP Fatal error: Class 'MigrateFieldHandler' not found
#2570239: Display date range as span instead of div
#2339119: Notice: Undefined index: date in date_text_element_process()
#2331975: Sync end date for the date popup functionality
#2026545: Date field's interval formatter always uses start date.
#2038057: date_select element's default value as an array raises a warning
#1874422: Pager may cause incorrect "prev" or "next" link
#1699312: Fatal error: Call to undefined function date_field_widget_settings_form_validate() in form.inc
#2532636: Undefined index: #default_value in date_popup_process_time_part()
#1504202: Allow entering of custom date format in Views field handlers
#1889744: PDOException when saving date as unix timestamp
#1355256: date field minutes and seconds rounding after update/save
#1797630: 'Is empty(null)' views filter is not working with exposed operators enabled when actual date values are left empty
#1889744: PDOException when saving date as unix timestamp
#2375235: Calendar block Next/Prev navigation broken
#1854220: Add module_exists so that error isn't thrown when disabling date_repeat_field module.
#2469189: Notice: Undefined index: show_remaining_days in date_field_formatter_view() (line 210 of …/date/date.field.inc).
#2325207: DateTimeZone could not be converted to string


tobybellwood:
closing in favour of #640

PR #562, opened 2017-10-05
Update file_entity and media to the latest stable releases to avoid i…

seanhamlin:
…ssues with duplicate files.

See Acquia tickets #410327, #499765, #584147


invisigoth (changes requested):
Both "file entity" and "media" modules have rolled out new releases since this PR that have addressed numerous issues in the versions in this PR. Suggest we revise this PR to use the latest release for both of those modules.

tobybellwood:
Closing in favour of PRs #591 and #592 to test and update seperately.

PR #566, opened 2017-10-10
Add Patch to Search API Pages module to remove unnecessary H1

deanvucic:
## https://www.drupal.org/node/2900795#comment-12224003

Patch Notes:
Since the page already has a H1 page title for these pages, we shouldn't be printing `<h1>Search</h1>` in the search-api-page-full-page.tpl.php file. Page title is already handled in page.tpl.php so the second H1 is redundant.


pandaskii:
Closed this PR in favor of https://github.com/govCMS/govCMS/pull/602

PR #582, opened 2017-10-25
Update feeds to 7.x-2.0-beta4 (from 7.x-2.0-beta3)

tobybellwood:
## https://www.drupal.org/project/feeds/releases/7.x-2.0-beta4

Release notes
Note
If you are having trouble running database updates, make sure that the Feeds module is enabled. See #2911711: FeedsHTTPCache class could not be found: 7.x-2.0-beta4 database update can only run if Feeds is enabled

In this release a big issue is fixed: downloaded content is no longer cached in the database! An other improvement is that you can now see on the import page when Feeds will run the next import. For this release there is also a small chance of a backwards compatibility break if you use a custom fetcher. Read on for more details.

Methods have been added to FeedsFetcherResult

The base class for fetcher results, FeedsFetcherResult, has had a few changes: methods have been added to this class and it now implements the magic method __sleep(). This means that this will break custom fetchers from which the fetcher result class has methods implemented with the same name. It can also break fetcher result classes that have private properties. For more information, see the change record.

Downloaded content is no longer cached in the database!

Yes! No more data size issues if you use a cache backend like memcache, for example. If you are using the default cache backend instead, your database will also grow less big.
Previously HTTP downloaded content was cached in the database and that could add up a lot of space, especially when you are fetching images via HTTP. This data is now cached on the file system instead, while the headers are still saved in the database. To keep the cached headers in the database in sync with the cached content on the file system, a cache class called 'FeedsHTTPCache' is added for the cache bin 'cache_feeds_http'.

Want to use a different cache class for the cache_feeds_http bin?
This is still possible. The cache logic in Feeds is made so that even when you would use a different cache class for the cache_feeds_http bin, the content is still cached on the file system. Be aware though that if you use a different cache class, cached files are not immediately cleaned up when you clear caches. To work around this problem, Feeds will check each six hours if there are orphaned files in the cache directory and cleans them up if there are (you can optionally in- or decrease the check interval by setting the 'feeds_sync_cache_feeds_http_interval' variable). Do not put other files in the Feeds cache directory, as Feeds will delete these!
See the change record for more information.

Next import time

An other improvement is that you can now see on the import page if and when the next import is scheduled. This information is handy if you are facing issues with imports not happening on cron. In such case, you can then at least see if the next import is properly scheduled.
You may still encounter issues with imports not happening during cron and one of these causes is planned to be fixed in the next release:
#2541944: Taxonomy mapping does not work with Cron

Also new in this release:

hook_feeds_config_defaults()
With this hook you can add additional configuration to the importer. Handy if you want to add some options via a form alter hook.
For the HTTP fetcher, you can optionally disable caching downloaded data.
Feeds will warn you if it detects invalid or obscure configuration for the following cases:
The content type/bundle selected on the processor no longer exists.
The language selected on the processor no longer exists.
The importer is attached to the same content type as the content type selected on the processor (a common mistake that newbies make).
Process in background is turned on, but import on submission is turned off (and using the standalone form). In this case, Feeds will do nothing when submitting the import form (other than saving the source).
On imports field validation checks are performed. This decreases the chance of SQL errors and invalid import data. In rare cases, it can also cause that some data is not being imported, while it imported without problems in previous releases (because the Field API now marks it as invalid).
Want to downgrade to 7.x-2.0-beta3?

There is a small chance that this release introduces changes that are not backwards compatible (see 'Methods have been added to FeedsFetcherResult' above). In such case you may want to go back to 7.x-2.0-beta3. This will cause the issue of the FeedsHTTPCache class not being found. To fix this:

Delete the variable 'cache_class_cache_feeds_http': DELETE FROM variable WHERE name = 'cache_class_cache_feeds_http';.
Truncate the table 'cache_bootstrap': TRUNCATE TABLE cache_bootstrap;.
Set the schema version of Feeds back to its previous value, so the cache class setting can be installed again at a later time: UPDATE system SET schema_version = 7212 WHERE name = 'feeds';
All changes since 7.x-2.0-beta3:

Issue #2911282 by MegaChriz: Fixed do not show bundle warning for non-entity processors.
Issue #2911281 by MegaChriz: changed messages related to rescheduling imports.
Issue #2879005 by MegaChriz, anup.singh: fixed array_flip() notice that occurred when mapping a float value to node author.
Issue #1962006 by MegaChriz, gordon, twistor: Added hook hook_feeds_config_defaults() to allow other modules to add extra configuration.
Issue #2884308 by ppavels: Fixed import links in admin menu for feed importers that are attached to a content type which name contains underscores.
Issue #1360910 by MegaChriz, leendertdb: Added an option for the HTTP Fetcher to disable caching content fetched via HTTP.
Issue #2877125 by MegaChriz: Replaced http_request_get() with feeds_http_request() in order to easier implement more options for HTTP requests.
Issue #2662892 by MegaChriz: Do not cache downloaded source statically to improve PHP memory usage.
Issue #2854548 by MegaChriz: Don't display a warning about "Attach to content type" setting being the same as the node processor's content type when the importer is not attached to a content type.
Issue #2854548 by MegaChriz: Added warning when user has the "Attach to content type" setting the same as the node processor's content type.
Issue #2445477 by MegaChriz: Added validation for combination of settings that causes the "Process in background" setting to have no effect.
Issue #2630694 by MegaChriz, twistor: Run background jobs directly in queue.
Issue #2868134 by MegaChriz: Show next time that the source will be imported.
Issue #2450365 by MegaChriz, twistor, littledynamo: Fixed importer is not rescheduled for new sources when using attach to node.
Issue #2871580 by MegaChriz: Fixed jobs for no longer existing feeds are kept being scheduled.
Issue #2871583 by MegaChriz: Fixed feeds jobs are not cleaned up when uninstalling Feeds.
Issue #2866377 by permanovd, MegaChriz: Fixed release file handle on exception.
Issue #2320781 by MegaChriz: Added validation for feed importer configuration: check for invalid bundle and invalid language.
Issue #2839706 by wmostrey, MegaChriz: Allow Feeds Importers to be edited even when disabled.
Issue #2867182 by MegaChriz: Added a finishImport() method to FeedsSource to bundle tasks that need to be done at the end of an import.
Issue #2867189 by MegaChriz: Added a method to FeedsSource to unlock a feed.
Issue #2087091 by MegaChriz: Fixed clear states properly after import.
Issue #2829097 by MegaChriz: fixed don't store raw source in feeds_source table.
Issue #2863209 by MegaChriz: fixed class 'FeedsHTTPCache' not found in _cache_get_object() during/after performing database updates.
Issue #2860047 by skessler, MegaChriz: Fixed link to feeds log page in watchdog call.
Issue #2847312 by mikran, MegaChriz: Added entity metadata properties for feeds_item GUID and URL.
Issue #2788125 by MegaChriz, PaulDinelle: Added support for i18n_node bundle language settings.
Issue #2857935 by MegaChriz: Fixed failing test FeedsMapperMultilingualFieldsTestCase::testChangedLanguageImportForExistingNode().
Issue #2854152 by renatog, MegaChriz: Delete feeds variables when uninstalling Feeds.
Issue #2851707 by MegaChriz: Added user validation.
Issue #2752661 by MegaChriz, geru: build the list of available sources and targets in feeds_ui_mapping_form() instead of in the theme function, so Feeds Tamper can alter the form to fix the bug of temporary targets not showing in the UI.
Issue #1037188 by saurabh.tripathi.cs, msti, MegaChriz, twistor: Log exceptions that are thrown when trying to fetch from a HTTP source.
Issue #2829096 by MegaChriz: Cache result of HTTP source on file system.
Issue #2828605 by David_Rothstein: fixed "The following module has moved within the file system" warning that occurred when downloading or enabling a new module or when enabling a previously installed module that was moved on the file system.
Issue #2762339 by MegaChriz: fixed feed node form inherits settings from the standalone importer when switching from standalone to attach to content type.
Issue #2839464 by MegaChriz: improved error message for when a download from a url fails.
Issue #2379631 by MegaChriz, cboyden, twistor, joelpittet, robhardwick, dsnopek: fixed field_attach_validate() must be called before programmatic entity saves.
Issue #2445997 by MegaChriz: fixed do not load all Feeds plugins in feeds_forms() when no Feeds form is requested.
Issue #2838283 by MegaChriz: Fixed failing test FeedsMapperDateMultipleTestCase::test().
Issue #2834026 by bburg, MegaChriz: Fixed do not add blank column headers in CSV template.


pandaskii:
There is an "issue" - database update can only run if Feeds is enabled https://www.drupal.org/node/2911711

invisigoth (commented):
After a careful review I have found this update is NOT compatible with feeds found on at least one major agency site that will result in the import process to fail completely on every import. The release notes offer a clue:

”In rare cases, it can also cause that some data is not being imported, while it imported without problems in previous releases (because the Field API now marks it as invalid). Want to downgrade to 7.x-2.0-beta3?”

Unfortunately we have sites falling into those "rare cases".

deanvucic:
Due to incompatibility with sites, I am closing this PR.

tobybellwood:
Reopening PR, possibility to unblock now memcache 1.6 is in place on ACSF

PR #585, opened 2017-11-15
Update Modernizr to 2.8.3 (from 2.7.0)

deanvucic:
## Release notes (v2.8)
https://github.com/Modernizr/Modernizr/releases/tag/v2.8.0

- Fix ie 10 form-validation detection - #1060
- Fix crash with Modernizr crashing firefox when used in a hidden iframe - #886
- Fix notification api test - #1015
- Improve websocket detect - 19402ed
- Update blob constructor test due to deprecation in Firefox - #1141
- Add css-shapes detect - #1188

## Release notes (v2.8.1)
https://github.com/Modernizr/Modernizr/releases/tag/v2.8.1

- fix css-shapes test

## Release notes (v2.8.2)
https://github.com/Modernizr/Modernizr/releases/tag/v2.8.2

- formvalidation was not backported correctly.

## Release notes (v2.8.3)
https://github.com/Modernizr/Modernizr/releases/tag/v2.8.3

- fix backporting of workers-blobworkers detect

PR #578, opened 2017-10-11
Update navbar to 7.x-1.7 (from 7.x-1.6)

deanvucic:
## https://www.drupal.org/project/navbar/releases/7.x-1.7

Release notes
Issue #2550367 by eshta: Use full window location for back to site length.
Issue #2571289 by eshta, hass: Rename 'Menu' to 'Manage' to better align with Drupal 8
Issue #2434655 by eshta, partdigital: Update menu tree build process to review access for each load
Issue #2509740 by eamonogealagain: Fix Modernizr version detection regex.
Issue #2477473 by eshta: In narrowest viewports, make the vertical toolbar menu 100% width
Issue #2448023 by eshta: Remove unused CSS.
Issue #2182939 by eshta: Add modernizr.min.js as an accepted minified filename.
Issue #2429669 by eshta: Fix body 'shake' on table pages by caching the top displacement amount.
Issue #2448925 by msound: Fix line ending in navbar.make.example.
Issue #2227293 by hass, lsolesen: Remove navbar from print view.

PR #588, opened 2017-11-20
Update Backbone to 1.3.3 (from 1.1.0) and Update Underscore to 1.8.3 (from 1.5.2)

deanvucic:
# Backbone Release Notes

## https://github.com/jashkenas/backbone/releases/tag/1.3.3
1.3.3 — Mar. 12, 2016
◦ Added findIndex and findLastIndex Underscore methods to Collection. 
◦ Added options.changes to Collection "update" event which includes added, merged, and removed models. 
◦ Ensured Collection#reduce and Collection#reduceRight work without an initial accumulator value. 
◦ Ensured Collection#_removeModels always returns an array. 
◦ Fixed a bug where Events.once with object syntax failed to bind context. 
◦ Fixed Collection#_onModelEvent regression where triggering a change event without a model would error. 
◦ Fixed Collection#set regression when parse returns a falsy value. 
◦ Fixed Model#id regression where id would be unintentionally undefined. 
◦ Fixed _removeModels regression which could cause an infinite loop under certain conditions. 
◦ Removed component package support. 

## https://github.com/jashkenas/backbone/releases/tag/1.2.3
1.2.3 — Sept. 3, 2015
◦ Fixed a minor regression in 1.2.2 that would cause an error when adding a model to a collection at an out of bounds index. 

## https://github.com/jashkenas/backbone/releases/tag/1.2.2
1.2.2 — Aug. 19, 2015
◦ Collection methods find, filter, reject, every, some, and partition can now take a model-attributes-style predicate: this.collection.reject({user: 'guybrush'}). 
◦ Backbone Events once again supports multiple-event maps (obj.on({'error change': action})). This was a previously undocumented feature inadvertently removed in 1.2.0. 
◦ Added Collection#includes as an alias of Collection#contains and as a replacement for Collection#include in Underscore.js >= 1.8. 

## https://github.com/jashkenas/backbone/releases/tag/1.2.1
1.2.1 — Jun. 4, 2015
◦ Collection#add now avoids trying to parse a model instance when passed parse: false. 
◦ Bug fix in Collection#remove. The removed models are now actually returned. 
◦ Model#fetch no longer parses the response when passing patch: false. 
◦ Bug fix for iframe-based History when used with JSDOM. 
◦ Bug fix where Collection#invoke was not taking additional arguments. 
◦ When using on with an event map, you can now pass the context as the second argument. This was a previously undocumented feature inadvertently removed in 1.2.0. 

## https://github.com/jashkenas/backbone/releases/tag/1.2.0
1.2.0 — May 13, 2015
◦ Added new hooks to Views to allow them to work without jQuery. See the wiki page for more info. 
◦ As a neat side effect, Backbone.History no longer uses jQuery's event methods for pushState and hashChange listeners. We're native all the way. 
◦ Also on the subject of jQuery, if you're using Backbone with CommonJS (node, browserify, webpack) Backbone will automatically try to load jQuery for you. 
◦ Views now always delegate their events in setElement. You can no longer modify the events hash or your view's el property in initialize. 
◦ Added an "update" event that triggers after any amount of models are added or removed from a collection. Handy to re-render lists of things without debouncing. 
◦ Collection#at can take a negative index. 
◦ Added modelId to Collection for generating unique ids on polymorphic collections. Handy for cases when your model ids would otherwise collide. 
◦ Added an overridable _isModel for more advanced control of what's considered a model by your Collection. 
◦ The success callback passed to Model#destroy is always called asynchronously now. 
◦ Router#execute passes back the route name as its third argument. 
◦ Cancel the current Router transition by returning false in Router#execute. Great for checking logged-in status or other prerequisites. 
◦ Added getSearch and getPath methods to Backbone.History as cross-browser and overridable ways of slicing up the URL. 
◦ Added delegate and undelegate as finer-grained versions of delegateEvents and undelegateEvents. Useful for plugin authors to use a consistent events interface in Backbone. 
◦ A collection will only fire a "sort" event if its order was actually updated, not on every set. 
◦ Any passed options.attrs are now respected when saving a model with patch: true. 
◦ Collection#clone now sets the model and comparator functions of the cloned collection to the new one. 
◦ Adding models to your Collection when specifying an at position now sends the actual position of your model in the add event, not just the one you've passed in. 
◦ Collection#remove will now only return a list of models that have actually been removed from the collection. 
◦ Fixed loading Backbone.js in strict ES6 module loaders. 

## https://github.com/jashkenas/backbone/releases/tag/1.1.2
1.1.2 — Feb. 20, 2014
◦ Backbone no longer tries to require jQuery in Node/CommonJS environments, for better compatibility with folks using Browserify. If you'd like to have Backbone use jQuery from Node, assign it like so: Backbone.$ = require('jquery'); 
◦ Bugfix for route parameters with newlines in them. 

## https://github.com/jashkenas/backbone/releases/tag/1.1.1
1.1.1 — Feb. 13, 2014
◦ Backbone now registers itself for AMD (Require.js), Bower and Component, as well as being a CommonJS module and a regular (Java)Script. Whew. 
◦ Added an execute hook to the Router, which allows you to hook in and custom-parse route arguments, like query strings, for example. 
◦ Performance fine-tuning for Backbone Events. 
◦ Better matching for Unicode in routes, in old browsers. 
◦ Backbone Routers now handle query params in route fragments, passing them into the handler as the last argument. Routes specified as strings should no longer include the query string ('foo?:query' should be 'foo'). 

# Underscore Release Notes

## https://github.com/jashkenas/underscore/releases/tag/1.8.3
1.8.3 — April 2, 2015

◦ Adds an _.create method, as a slimmed down version of Object.create. 
◦ Works around an iOS bug that can improperly cause isArrayLike to be JIT-ed. Also fixes a bug when passing 0 to isArrayLike. 

## https://github.com/jashkenas/underscore/releases/tag/1.8.2
1.8.2 — Feb. 22, 2015 — Diff — Docs

◦ Restores the previous old-Internet-Explorer edge cases changed in 1.8.1. 
◦ Adds a fromIndex argument to _.contains. 

## https://github.com/jashkenas/underscore/releases/tag/1.8.1
1.8.1 — Feb. 19, 2015 — Diff — Docs

◦ Fixes/changes some old-Internet Explorer and related edge case behavior. Test your app with Underscore 1.8.1 in an old IE and let us know how it's doing... 

## https://github.com/jashkenas/underscore/releases/tag/1.8.0
1.8.0 — Feb. 19, 2015 — Diff — Docs

◦ Added _.mapObject, which is similar to _.map, but just for the values in your object. (A real crowd pleaser.) 
◦ Added _.allKeys which returns all the enumerable property names on an object. 
◦ Reverted a 1.7.0 change where _.extend only copied "own" properties. Hopefully this will un-break you — if it breaks you again, I apologize. 
◦ Added _.extendOwn — a less-useful form of _.extend that only copies over "own" properties. 
◦ Added _.findIndex and _.findLastIndex functions, which nicely complement their twin-twins _.indexOf and _.lastIndexOf. 
◦ Added an _.isMatch predicate function that tells you if an object matches key-value properties. A kissing cousin of _.isEqual and _.matcher. 
◦ Added an _.isError function. 
◦ Restored the _.unzip function as the inverse of zip. Flip-flopping. I know. 
◦ _.result now takes an optional fallback value (or function that provides the fallback value). 
◦ Added the _.propertyOf function generator as a mirror-world version of _.property. 
◦ Deprecated _.matches. It's now known by a more harmonious name — _.matcher. 
◦ Various and diverse code simplifications, changes for improved cross-platform compatibility, and edge case bug fixes. 

## https://github.com/jashkenas/underscore/releases/tag/1.7.0
1.7.0 — August 26, 2014 — Diff — Docs

◦ For consistency and speed across browsers, Underscore now ignores native array methods for forEach, map, reduce, reduceRight, filter, every, some, indexOf, and lastIndexOf. "Sparse" arrays are officially dead in Underscore. 
◦ Added _.iteratee to customize the iterators used by collection functions. Many Underscore methods will take a string argument for easier _.property-style lookups, an object for _.where-style filtering, or a function as a custom callback. 
◦ Added _.before as a counterpart to _.after. 
◦ Added _.negate to invert the truth value of a passed-in predicate. 
◦ Added _.noop as a handy empty placeholder function. 
◦ _.isEmpty now works with arguments objects. 
◦ _.has now guards against nullish objects. 
◦ _.omit can now take an iteratee function. 
◦ _.partition is now called with index and object. 
◦ _.matches creates a shallow clone of your object and only iterates over own properties. 
◦ Aligning better with the forthcoming ECMA6 Object.assign, _.extend only iterates over the object's own properties. 
◦ Falsey guards are no longer needed in _.extend and _.defaults—if the passed in argument isn't a JavaScript object it's just returned. 
◦ Fixed a few edge cases in _.max and _.min to handle arrays containing NaN (like strings or other objects) and Infinity and -Infinity. 
◦ Override base methods like each and some and they'll be used internally by other Underscore functions too. 
◦ The escape functions handle backticks (`), to deal with an IE ≤ 8 bug. 
◦ For consistency, _.union and _.difference now only work with arrays and not variadic args. 
◦ _.memoize exposes the cache of memoized values as a property on the returned function. 
◦ _.pick accepts iteratee and context arguments for a more advanced callback. 
◦ Underscore templates no longer accept an initial data object. _.template always returns a function now. 
◦ Optimizations and code cleanup aplenty. 

## https://github.com/jashkenas/underscore/releases/tag/1.6.0
1.6.0 — February 10, 2014 — Diff — Docs

◦ Underscore now registers itself for AMD (Require.js), Bower and Component, as well as being a CommonJS module and a regular (Java)Script. An ugliness, but perhaps a necessary one. 
◦ Added _.partition, a way to split a collection into two lists of results — those that pass and those that fail a particular predicate. 
◦ Added _.property, for easy creation of iterators that pull specific properties from objects. Useful in conjunction with other Underscore collection functions. 
◦ Added _.matches, a function that will give you a predicate that can be used to tell if a given object matches a list of specified key/value properties. 
◦ Added _.constant, as a higher-order _.identity. 
◦ Added _.now, an optimized way to get a timestamp — used internally to speed up debounce and throttle. 
◦ The _.partial function may now be used to partially apply any of its arguments, by passing _ wherever you'd like a placeholder variable, to be filled-in later. 
◦ The _.each function now returns a reference to the list for chaining. 
◦ The _.keys function now returns an empty array for non-objects instead of throwing. 
◦ … and more miscellaneous refactoring. 

## https://github.com/jashkenas/underscore/releases/tag/1.5.2
1.5.2 — September 7, 2013 — Diff — Docs

◦ Added an indexBy function, which fits in alongside its cousins, countBy and groupBy. 
◦ Added a sample function, for sampling random elements from arrays. 
◦ Some optimizations relating to functions that can be implemented in terms of _.keys (which includes, significantly, each on objects). Also for debounce in a tight loop. 
◦ The _.escape function no longer escapes '/'. 

## https://github.com/jashkenas/underscore/releases/tag/1.5.1
1.5.1 — July 8, 2013 — Diff — Docs

◦ Removed unzip, as it's simply the application of zip to an array of arguments. Use _.zip.apply(_, list) to transpose instead. 

## https://github.com/jashkenas/underscore/releases/tag/1.5.0
1.5.0 — July 6, 2013 — Diff — Docs

◦ Added a new unzip function, as the inverse of _.zip. 
◦ The throttle function now takes an options argument, allowing you to disable execution of the throttled function on either the leading or trailing edge. 
◦ A source map is now supplied for easier debugging of the minified production build of Underscore. 
◦ The defaults function now only overrides undefined values, not null ones. 
◦ Removed the ability to call _.bindAll with no method name arguments. It's pretty much always wiser to white-list the names of the methods you'd like to bind. 
◦ Removed the ability to call _.after with an invocation count of zero. The minimum number of calls is (naturally) now 1. 

## https://github.com/jashkenas/underscore/releases/tag/1.4.4
1.4.4 — January 30, 2013 — Diff — Docs

◦ Added _.findWhere, for finding the first element in a list that matches a particular set of keys and values. 
◦ Added _.partial, for partially applying a function without changing its dynamic reference to this. 
◦ Simplified bind by removing some edge cases involving constructor functions. In short: don't _.bind your constructors. 
◦ A minor optimization to invoke. 
◦ Fix bug in the minified version due to the minifier incorrectly optimizing-away isFunction. 

## https://github.com/jashkenas/underscore/releases/tag/1.4.3
1.4.3 — December 4, 2012 — Diff — Docs

◦ Improved Underscore compatibility with Adobe's JS engine that can be used to script Illustrator, Photoshop, and friends. 
◦ Added a default _.identity iterator to countBy and groupBy.


deanvucic:
Backbone v1.3.3 is dependent on Underscore v1.8.3. These libraries are used by the NavBar module and possibly breakpoint which is dependent on Navbar.

courdo1 (commented):
Two open bugs that effect underscore 1.8.3
https://github.com/jashkenas/underscore/issues/2214;
https://github.com/jashkenas/underscore/issues/1834

One open issue with backbone 1.3.3
https://github.com/jashkenas/backbone/issues/4159


These bugs are not show stoppers.

courdo1 (approved):
ready for testing

pandaskii:
For testing, please hold until all these PRs are approved

https://github.com/govCMS/govCMS/pull/578

#588

#585

@Venkata-modgu @deanvucic

Venkata-modgu:
Navbar tested in conjunction with the Backbone, Underscore and Modernizr library updates. 
Navigation bar working fine. 
Test passed.

PR #594, opened 2017-11-27
Update entitycache to 7.x-1.5 (from 7.x-1.2)

deanvucic:
## https://www.drupal.org/project/entitycache/releases/7.x-1.5

Release notes
#2441965 Broken updates due to Classes living in .module files

## https://www.drupal.org/project/entitycache/releases/7.x-1.4

Release notes
Changes since 7.x-1.3:

#1851398 Invalid text representation when attempting to load an entity with a string ID

## https://www.drupal.org/project/entitycache/releases/7.x-1.3

Release notes

#2441965 Broken updates due to Classes living in .module files
#1349566 Add tests for content translation
#1262430 Convert $conditions to use EntityFieldQuery then load by ID from cache
#2135169 Support caching of failed load operations.
#1670848 Fixed Unnecessary recaching of terms when using taxonomy_get_term_by_name.
#2086633 Allow for the clearing of individual entitycache caches via drush cc


deanvucic:
Updates from 7.x-1.2 may fail on some sites with fatal errors due to issues with Drupal 7's classloader, making update.php and the rest of the site inaccessible.

There is no way to resolve this in entitycache itself, so it requires an additional step when updating the module.

Ensure you have drush available for your install, or download https://www.drupal.org/project/registry_rebuild and follow the instructions before updating from 7.x-1.2 to 7.x-1.5.

Then when you update, rebuild the registry and everything should be fine again.

https://www.drupal.org/project/registry_rebuild

PR #602, opened 2017-11-27
Add Patch to remove double H1 page titles

deanvucic:
Since the page already has a H1 page title for these pages, we shouldn't be printing `<h1>Search</h1> ` in the search-api-page-full-page.tpl.php file.

Page title is already handled in page.tpl.php so the second H1 is redundant.

This patch fixes duplicated H1 page titles.
image

Conversation at https://community.govcms.gov.au/t/duplicate-search-page-h1-and-breadcrumbs-in-out-of-the-box-govcms-distribution/864/7


pandaskii (approved):
This patch is already in the module dev branch.

Approved

Ready for testing

Venkata-modgu:
Test passed.

PR #612, opened 2017-11-28
Update token to 7.x-1.7 (from 7.x-1.6)

deanvucic:
## https://www.drupal.org/project/token/releases/7.x-1.7

Release notes
Changes since 7.x-1.6:

#2825841 by hbemtec: Fixed PHP 7.1 compatibility error: Cannot use string offset as an array in _token_token_tree_format_row().
#2702379 by Dave Reid: Review @todos in token_element_validate()
#2630378 by hussainweb: Token "cannot contain fewer than one token" error misleading
#2474403 by kfritsche, mariamania, Zekvyrin, B-Prod: Fixed token replacement of field descriptions should not replace i18n_field.module's translations.
#2485221 by Berdir: Fixed token_pre_render_field_token() parameter should not be passed by reference.
#2299039 by Elijah Lynn: Remove unused $sanitize variable from field_tokens().
Soften token status report messages, split into individual items with different severity values.
Fixed tokens without a description value were treated as wrong.
#1821902 by Antonin Brunon, emattias: Fixed palse positives in recursive token check in _token_build_tree() for tokens that had a name of '0' (e.g. entity_token module's delta '0' field tokens).


deanvucic:
The token browser fails to load on PHP7.1 (admin/config/search/path/patterns -> 'browse available tokens'). The error is not immediate apparent unless one manually sifts through the drupal log messages (admin/reports/dblog). Updating the module to 1.7 resolves this issue.

PR #629, opened 2018-01-18
Added default security kit configuration

courdo1:
Added default security kit configuration


tobybellwood:
Blocked - not wanting to set these overly assertive settings at the distribution.  Perhaps recommend them?

tobybellwood:
Not proceeding - too prescriptive to set this at a distro level.

PR #616, opened 2017-11-29
Update webform to 7.x-4.16 (from 7.x-4.15)

deanvucic:
Of particular interest here is the "upload destination private files by default if available".

## https://www.drupal.org/project/webform

Changes since 7.x-4.15:

Coding standards
#2816303 by ilari.stenroth, Liam Morland, vinmassaro: Make file upload destination private files by default if available
#2901156 by dhruveshdtripathi, raynimmo: Fix undefined index: #id in theme_webform_element()
#2900109 by lambic: Undefined function webform_component_feature() when entitycache is enabled
#2713743 by joelstein: Add @placeholder support to Number component
#2604830 by heykarthikwithu, joelpittet: Code cleanup: Remove unused variables throughout the module
#2883514: Coding standards
#2873783: Disable form controls when hiding them
#2869746 by torotil: Return render arrays instead of strings
#2780283 by nightwalkr, Liam Morland: Fix undefined index in webform_component_include()
#2866630 by alexp999, Liam Morland: PHP Zip requirement should be Info, not Warning
#2683801 by nicrodgers, joel_osc: Translation of progress bar page breaks
#2857854 by pifagor, Liam Morland: Fix non-numeric value warning in PHP 7.1
#2860487 by mercepedraza: Set numeric values as "general" in Excel export
#2867012 by Munavijayalakshmi: Fix documentation spelling errors

PR #632, opened 2018-02-01
Add swiftmailer 7.x-1.7 and library v5.4.9

tobybellwood:
_Assessing suitability of swiftmailer over mimemail to provide solution for emailing private files without requiring excessively broad permissions._

## https://www.drupal.org/project/swiftmailer

The Swift Mailer module extends the basic e-mail sending functionality provided by Drupal by delegating all e-mail handling to the Swift Mailer library. This enables your site to take advantage of the many features which the Swift Mailer library provides, such as :

- Sending e-mails directly through a SMTP server of your choice, a locally installed MTA agent such as sendmail or the mail functionality provided by PHP.
- Sending HTML (MIME) e-mails.
- Adding file attachments to e-mails.
- Adding inline images to e-mails.

The module also lets you theme e-mails so that they reflect the general look and feel of your site and at the same time appear correctly in various e-mail clients. Images required by the theme can be attached to the e-mail and displayed inline.

The module is also capable of automatically creating a plain text version based on the HTML version of an e-mail.

The Swift Mailer module depends on the mailsystem module. This module is responsible for actually making the Swift Mailer module available to Drupal. Furthermore, you are also required to download the Swift Mailer library to get up and running with this module.

## ## https://github.com/swiftmailer/swiftmailer

Using v5.4.9 as latest with PHP 5.x compatibility


tobybellwood:
I think we may struggle given the need for composer install - under http://632.govcms.xyz/admin/config/people/swiftmailer the library _appears_ to be installed in profiles/govcms/libraries/swiftmailer - but currently unable to test properly

tobybellwood:
Added to d.o. packaging whitelist in case we go this way instead of #633 (https://www.drupal.org/packaging-whitelist?title=swift)

invisigoth (approved):
Reviewed the following:
- Source code
- Change log
- Issue queue
It appears the 7.x branch of this module is stable and is suitable to be included in the distribution.

tobybellwood:
Testing to include:

Set swiftmailer as default mailsystem (instead of mimemail).
Ensure system emails are properly received.
Tests with #616 (details there)

PR #635, opened 2018-02-10
Added in php71 patch to fix navbar

tobybellwood:
Adding in patch from https://www.drupal.org/project/libraries/issues/2779591

> If `$library['version callback']` uses a pass by reference parameter (and does not have `$library['version arguments'][0]>` set, then a warning is thrown in a PHP 7.0 environment. Issue can cause referenced library not to load correctly. This was specifically seen with the Navbar module which used the function `_navbar_libraries_get_version(&$library, $options = array()).`
> 
> Example of the error:
> 
> > Warning: Parameter 1 to _navbar_libraries_get_version() expected to be a reference, value given in libraries_detect() (line 549 of libraries.module).
> 
> That particular line uses `call_user_func()` which only passes parameters by value and never by reference. It needs to be changed to `call_user_func_array()`, which is actually used as the other version callback function for a slightly different branch scenario.

For future reference patch-3 applies against 7.x-2.3 release, patch-11 against 7.x-2.x, but there is no pending release.


petrillek (approved):
I've tested this patch on last GovCMS release and it is working as expected. Haven't found any issues from that patch. We should merge it in, as the warning is anoying. :)

PR #639, opened 2018-02-20
Update bigmenu to 7.x-1.x (from 7.x-1.3)

deanvucic:
## https://cgit.drupalcode.org/bigmenu/log/
**Changes:**
- Follow up additional code clean-ups identified by PHP CodeSniffer
- Code Standard Clean ups
- Fix: Error when going to admin/structure/bigmenu-customize/main-menu


Feng-Shui:
I can't that any of the issues in log above are PHP 7.1 related.

The two code tidy commits don't seem to be 7.1 related, nor does the error when hitting that admin path issue, that's just removing an old admin path and getting the menu router to rebuild.

Am I missing something here?

As an aside, there's been no more commits against the dev branch, so the log above is still current.

PR #640, opened 2018-02-20
Update date to 7.x-2.x (from 7.x-2.9)

deanvucic:
## https://cgit.drupalcode.org/date/log/
**Changes / Fixes:**
- Fix PHP 7 tests
- String Offset errors on edit page using PHP 7.1.0
- 'clone' is a reserved keyword introduced in PHP version 5.0 and cannot be invoked as a function.
- Remove the .gitignore file
- Tidy up the CHANGELOG.txt file.
- Remove unnecessary space in data_combo title
- Missing help text
- DateMigrateExampleUnitTest fails if migrate is missing
- Tests are broken on current 7.x-2.x branch
- As per Drupal coding standards, use uppercase for PHP constants
- DateTime::format(): The DateTime object has not been correctly initialized by its constructor in DateObject->format()
- date_popup element with just time doesn't work
- Grammar issue: "Date objects can be created and cached as date fields are loaded rather than when they are displayed to improve performance"
- PHP Fatal error: Class 'MigrateFieldHandler' not found
- Display date range as span instead of div
- Notice: Undefined index: date in date_text_element_process()
- Sync end date for the date popup functionality
- Date field's interval formatter always uses start date
- date_select element's default value as an array raises a warning
- Pager may cause incorrect "prev" or "next" link
- Fatal error: Call to undefined function date_field_widget_settings_form_validate() in form.inc
- Undefined index: #default_value in date_popup_process_time_part()
- Allow entering of custom date format in Views field handlers
- PDOException when saving date as unix timestamp
- date field minutes and seconds rounding after update/save
- 'Is empty(null)' views filter is not working with exposed operators enabled when actual date values are left empty
- PDOException when saving date as unix timestamp
- Calendar block Next/Prev navigation broken
- Add module_exists so that error isn't thrown when disabling date_repeat_field module
- Undefined index: show_remaining_days in date_field_formatter_view() (line 210 of …/date/date.field.inc)
- DateTimeZone could not be converted to string


pandaskii:
Closed in favour #801

PR #619, opened 2017-11-30
Update views_bulk_operations to 7.x-3.5 (from 7.x-3.3)

deanvucic:
## https://www.drupal.org/project/views_bulk_operations/releases/7.x-3.5
Release notes
Changes since 7.x-3.4:

#2946220 by bkosborne: PHP 7.2+ future proofing, remove usage of deprecated "create_function()"
#2334625 by g.oechsler, larowlan: "Select all items on all pages" executed from last page affects only last page items
#2674166 by hbemtec, nagy.balint, joelpittet: "Select all" checkbox and sticky headers
#2475873 by attiks, joelpittet: Javascript blocks clicks on label elements
#2890878 by matthensley: Decrease js specificity for use with Bootstrap theme
#2800227 by joelpittet, cudevdev: 'Move to book' populates the menu_links table but not the book table.
#2718223 by joelpittet, Yuri: Fatal error: Call to undefined function EntityDefaultMetadataController::bundleOptionsList()
#2431131 by joelpittet, vijithaepa: Fatal error: Unsupported operand types in ...modules\views_bulk_operations\plugins\operation_types\action.class.php
#2873171 by joelpittet: PHP7 readiness
#2872389 by tmin: "Select All" option does not show up in some cases
#2867407 by hgoto: 2 patterns of jQuery selector are used for the same elements in views_bulk_operations.js
#2855939 by hgoto, stefanos.petrakis: Bulk operation buttons don't work after upgrade to 3.4 when Force Single selected
#2862687 by xdevx, joelpittet: Warning: call_user_func() expects parameter 1 to be a valid callback
#2855782 by stefanos.petrakis, hgoto, joelpittet, Liu Xin: Disable submit button works only after a second selection is done: Recent jQuery+WholeRowClickable
#2860320 by claudiu.cristea, joelpittet: Allow exceptions when toggling buttons
#2856575 by plach: "Select all on all pages" widget always displayed when the view has no pager
#2853029 by joelpittet, pjohn, tobiasb, nairb: views_bulk_operations_action_load_list not working
#2473889 by CvW: rules component with list as parameter creates a notice

## https://www.drupal.org/project/views_bulk_operations/releases/7.x-3.4

Release notes
Changes since 7.x-3.3:

#2505457 by dago.aceves, sambonner, joelpittet, joekers, k2abhishek, gbyte.co, CHENZEN002, klidifia, edward.radau: Cancel User Account: Delete account and make content belong to "Anonymous" issue
[...]

PR #645, opened 2018-02-21
Update views_bulk_operations to 7.x-3.x (from 7.x-3.3)

deanvucic:
## https://cgit.drupalcode.org/views_bulk_operations/commit/?id=d7638c85e03e4348ab8296a5f19b33a0411fd11f

**Release notes**
_Changes since 7.x-3.4:_
- PHP7 readiness
- rules component with list as parameter creates a notice
- views_bulk_operations_action_load_list not working
- "Select all on all pages" widget always displayed when the view has no pager
- Allow exceptions when toggling buttons
- Disable submit button works only after a second selection is done: Recent jQuery+WholeRowClickable
- Warning: call_user_func() expects parameter 1 to be a valid callback
- Bulk operation buttons don't work after upgrade to 3.4 when Force Single selected
- 2 patterns of jQuery selector are used for the same elements in views_bulk_operations.js
- "Select All" option does not show up in some cases
- Unsupported operand types in ...modules\views_bulk_operations\plugins\operation_types\action.class.php
- Call to undefined function EntityDefaultMetadataController::bundleOptionsList()
- 'Move to book' populates the menu_links table but not the book table.
- Decrease js specificity for use with Bootstrap theme
- Javascript blocks clicks on label elements
- "Select all" checkbox and sticky headers
- "Select all items on all pages" executed from last page affects only last page items

## https://www.drupal.org/project/views_bulk_operations/releases/7.x-3.4
**Release notes**
_Changes since 7.x-3.3:_
- Cancel User Account: Delete account and make content belong to "Anonymous" issue
- Disable submit button until a selection is made
- VBO operation fails via Drush, succeeds via UI
- Disable submit button until a selection is made
- Load a list of entity from a VBO View doesn't work (AJAX error)
- Execute as User 1 - skip permission step
- Coding standard updates
- Warning: mb_strlen() expects parameter 1 to be string, array given in drupal_strlen()
- Archive files corrupted when they are in private file system
- "Select all" checkbox works only once
- [Warning] call_user_func() expects parameter 1 to be a valid callback, function 'views_bulk_operations_change_owner_action_info' not found or invalid function name views_bulk_operations.module:23
- Add action to change entity author
- Incorrect displaying of a strong tag on the confirmation step on the bootstrap theme
- Overwrites latest node revision
- Remove !important from CSS
- Re-use generic entity views table
- Fix all VBO views on page using first VBO field settings
- No watchdog entry created on bulk node deletion
- VBO expired queues aren't cleaned by cron


Feng-Shui:
VBO has gone to a 3.5 release, which includes PHP 7 (7.2) support.  @deanvucic any objections if this PR gets bumped to the 3.5 release?

tobybellwood:
covered by PR #619

PR #646, opened 2018-02-21
SA-CORE-2018-001 - Drupal core - Critical - Multiple Vulnerabilities

tobybellwood:
## https://www.drupal.org/sa-core-2018-001

Project: Drupal core
Version: 8.4.x-dev, 7.x-dev
Date: 2018-February-21
Security risk: Critical 16∕25 AC:Basic/A:User/CI:Some/II:Some/E:Exploit/TD:Default
Vulnerability: Multiple Vulnerabilities
Description: This security advisory fixes multiple vulnerabilities in both Drupal 7 and Drupal 8. See below for a list.

**JavaScript cross-site scripting prevention is incomplete - Critical - Drupal 7 and Drupal 8**
Drupal has a Drupal.checkPlain() JavaScript function which is used to escape potentially dangerous text before outputting it to HTML. This function does not correctly handle all methods of injecting malicious HTML, leading to a cross-site scripting vulnerability under certain circumstances.

The PHP functions which Drupal provides for HTML escaping are not affected.

**Private file access bypass - Moderately Critical - Drupal 7**
When using Drupal's private file system, Drupal will check to make sure a user has access to a file before allowing the user to view or download it. This check fails under certain conditions in which one module is trying to grant access to the file and another is trying to deny it, leading to an access bypass vulnerability.

This vulnerability is mitigated by the fact that it only occurs for unusual site configurations.

**jQuery vulnerability with untrusted domains - Moderately Critical - Drupal 7**
A jQuery cross site scripting vulnerability is present when making Ajax requests to untrusted domains. This vulnerability is mitigated by the fact that it requires contributed or custom modules in order to exploit.

For Drupal 7, it is fixed in the current release (Drupal 7.57) for jQuery 1.4.4 (the version that ships with Drupal 7 core) as well as for other newer versions of jQuery that might be used on the site, for example using the jQuery Update module.

**External link injection on 404 pages when linking to the current page - Less Critical - Drupal 7**
Drupal core has an external link injection vulnerability when the language switcher block is used. A similar vulnerability exists in various custom and contributed modules. This vulnerability could allow an attacker to trick users into unwillingly navigating to an external site.


pandaskii (changes requested):
Please remove the entity module update from drupal-org.make and keep only the core update

pandaskii (approved):

pandaskii:
We did review the changes of drupal.js and file.module etc, happy to merge if automation testing has no issues

Feng-Shui:
We've updated core this morning and testing on our site has shown no issues.

tobybellwood:
There is minimal issue-specific testing involved at this stage - govCMS has minimal exposure to these vulnerabilities.

General site functionality testing to be performed.

Venkata-modgu:
Test passed.

PR #647, opened 2018-02-22
Update govcms_social_links module to resolve PHP7.1 compatibility issue

deanvucic:
PHP7 no longer assumes undefined/empty string variables as arrays and needs to be explicitly defined. After upgrading to PHP7 from PHP5, the block contents do not render and returns the following PHP warning: "Illegal string offset in govcms_social_links_views"

To resolve this, the $block variable needs to be defined as an array() and not as an empty string.


tobybellwood:
Test this

PR #656, opened 2018-03-02
SA-CORE-2018-003 - Update CKEditor to 4.9.2 (from 4.5.9)

tobybellwood:
## https://www.drupal.org/sa-core-2018-003

CKEditor, a third-party JavaScript library included in Drupal core, has fixed a cross-site scripting (XSS) vulnerability. The vulnerability stemmed from the fact that it was possible to execute XSS inside CKEditor when using the image2 plugin (which Drupal 8 core also uses).

We would like to thank the CKEditor team for patching the vulnerability and coordinating the fix and release process, and matching the Drupal core security window.

Solution: 

- If you installed CKEditor in Drupal 7 using another method (for example with the WYSIWYG module or the CKEditor module with CKEditor locally) and you’re using a version of CKEditor from 4.5.11 up to 4.9.1, update the third-party JavaScript library by downloading CKEditor 4.9.2 from CKEditor's site.

We don't currently run >4.5.11 (so aren't actually affected), but it would probably be good practice to address it, if possible?

As per comments in Community Slack

> For the last couple of d7 sites of mine, I used CKEditor v4.8. You get the nag about wysiwyg only supporting up to version 4.6, But I've had no troubles with the later version. The Apply simple source formatting checkbox actually does what it says it should. Even the wysiwyg-templates module works with it too.


tobybellwood:
Note: 4.9.2 is outside supported range https://www.drupal.org/docs/7/modules/wysiwyg/supported-editors-matrix

Some conversation at https://www.drupal.org/project/wysiwyg/issues/2962192 around compatibility (and the source of the vulnerability)

ruwanl (approved):
Approved for testing

Venkata-modgu:
Test passed.

PR #677, opened 2018-04-19
[7.x-2.x] Updated funnelback module to 7.x-4.0-beta2

rujiali:
Update funnelback module to a more stable version 4.0-beta2.


tobybellwood:
Ok to test

mgdhs:
When I looked a couple of years ago the 2.x alpha was significantly different to 1.x. If a site currently uses 1.x I assume upgrading to 4.x will break their site search without additional steps?

Is anyone using this module in govCMS? 1.x was so broken and didn't work with a recent Funnelback API. I ended up hacking changes to the 1.x module to get basic search functionality working.

Edit: I feel a little bit silly considering @rujiali is the author of 4.x

rujiali:
I'm wondering if anyone is using Funnelback from govCMS distro as well. Since 1.x is broken anyway, I would assume anyone who is using it would have hacked it a little to make it work (agreed @mgdhs ). 4.x reserved the basic settings and the URL from 1.x, it would not break the basic functions from 1.x. But it does require user to check their configurations in Drupal and Funnelback end.

tobybellwood:
we'll be working with @rujiali to make sure any necessary changes are well documented.  I've had a runthrough of the improvements - and they look pretty neat!

tobybellwood:
Test this

PR #710, opened 2018-08-13
Remove google_analytics_reports module from the distribution dependencies

pandaskii:
Don't think it is a good idea to read the Google analytics reports from the site backend. However, we will keep it in the distribution for legacy reasons.

Limits and Quotas on API Requests https://developers.google.com/analytics/devguides/reporting/core/v3/limits-quotas#core_reporting


tobybellwood:
This has been added to the list of modules that are removed in the migration process - it's now ok to remove from distro.

PR #713, opened 2018-08-19
Adds search_api_multi module

pameeela:
Proposing to add search_api_multi module to allow searching across multiple indexes. My use case is wanting to index managed files (not the file contents, just the metadata) along with content. We couldn't think of another way besides possibly creating a node type for files, but that approach will 1) make our content migration much more complicated and 2) add a level of abstraction to files that will complicate content management forever.


tobybellwood:
H/T to @gargsuchi - To index two node types in one index, head to 
`admin/config/search/search_api/add_index` and select `Multiple types`

image

PR #717, opened 2018-08-27
Changed future date to be in the future again

tobybellwood:
Tests failing because we'd passed the future date in the test scenario


pandaskii (approved):
Super quick 👍

PR #719, opened 2018-08-30
Use xmlsitemap 7.x-2.5 stable release (instead of 7.x-2.x-dev)

tobybellwood:
2.5 tag at same commit as previous version


pandaskii (approved):
Thanks

PR #723, opened 2018-09-24
Make deployment branch configurable.

thom8:
Sets `DEPLOY_BRANCH` to `7.x-3.x`.

This branch will automatically push new beta images on every commit. 

These can be released with `ahoy release`.


tobybellwood (approved):
Good pickup - we don't use master, so all commits to 7.x-3.x now deploy to @beta , which is the correct configuration.  Tagged releases generate @latest for docker

PR #732, opened 2018-10-18
Remove unneeded modules (Acquia and GA Reports)

tobybellwood:
These are removed in the default-sanitisation/3_remove_acquia.modules process in migration


tobybellwood:
To be merged and assessed in conjunction with https://github.com/govCMS/govcmslagoon/pull/32.

- This PR removes the modules and as dependencies from the distribution as they are no longer needed.
- The govcmslagoon PR adds them back into the acsf image so they can be properly removed in a migration step.

govcms.install
```diff
@@ -128,8 +128,6 @@ function _govcms_create_basic_roles_perms() {
   );
 
   $extra_perms = array(
-    'administer google analytics reports',
```
ivangrynenko, line 131: Since the `$extra_perms` is an empty array now, maybe it worth cleaning this variable out and removing it from the  `$extended_permissions` below?

ivangrynenko (approved):
Apart from the minor cleanup suggestion, this PR looks okay.

mgdhs:
Please make sure this change is made clear in the release notes and announcement. A few GovCMS PaaS sites are remaining on Acquia that use the distribution. It shouldn't be an issue, but we'll need to be aware of it.

tobybellwood:
@mgdhs - we will make sure that the upgrade path from GovCMS 7.x-2.x to 7.x-3.x is clearly conveyed.  We created the 3.x branch exactly to make some of these long-needed decisions, as well as instigate a clear-out of sorts!

PR #738, opened 2018-10-31
Don't fail if amazeeio-network exists

tobybellwood:
If amazeeio-network already exists (such as running circleci build locally) the process will fail.

Added a `|| true` as per https://stackoverflow.com/questions/48643466/docker-create-network-should-ignore-existing-network


stooit (approved):
👍 
Same change already implemented for GovCMS8.

PR #756, opened 2018-12-06
Deploy tagged images on tag builds.

thom8:
There has been a change where `CIRCLE_BRANCH` is no longer set on tagged builds.

PR updates deployment conditions to deploy images on tagged builds.


pandaskii:
Merged in #774

PR #761, opened 2018-12-17
[GOVCMS-2193]: Optimisation for workbench_moderation in large volumes of revisions

fubarhouse:
Patch workbench_moderation to improve performance for nodes with large histories.


gargsuchi (approved):
I have queued testing for the patch.

PR #762, opened 2018-12-17
[GOVCMS-2012]: Support mobile phone numbers prefixed with 0499

fubarhouse:


gargsuchi (approved):

gargsuchi (changes requested):
@fubarhouse 
why don't we upgrade workbench module to 7.x-3.0 instead of applying this patch? The project page also mentions this:

* The 7.x-3.x branch is the recommended version and the successor of the 7.x-1.x branch. If you're using a 7.x-1.x version you should consider updating to the 7.x-3.x as it fixes following security issue: SA-CONTRIB-2016-060

fubarhouse:
Previous discussions determined we will not be upgrading to the specified version, so the patch is the only path forward.

gargsuchi (approved):

PR #770, opened 2019-01-16
Updates Drupal core to 7.63

fubarhouse:


fubarhouse:
@gargsuchi,

I've successfully tested this one without any issues. The tests were primitive but it does work without any unexpected results or messages.

Could you please review this one, and I'll continue on getting the pipeline to pass.

PR #777, opened 2019-01-30
[GOVCMSD7-28]: Updates govcms_account_security_final_validation

fubarhouse:
This change will resolve COVCMSD7-28 and update the function forcing user login validation to end on the specified callback to prevent other callback handlers from modifying the input before final validation occurs.


gargsuchi (approved):

fubarhouse:
Approval was provided internally.

PR #852, opened 2019-03-06
[GOVCMSD7-5] Update the patch file for phone for Australian phone number validation.

fubarhouse:
* [x] Follow-up to #776

This fixes the validation rules changed in #776, where the validation does not check the maximum length of a given validation rule, and the land-line validation length is incorrect.


fubarhouse:
Testing passed two hours ago by Venkata.

Note that another patch may be coming to prevent the use of non-numerical characters.

fubarhouse:
Based on feedback, there's now a mechanism to check for numeric values and fail when a string using other character types is inputted.

fubarhouse:
For the record this has been tested quite extensively - please see the internal ticketing system for more details.

ruwanl (approved):

PR #855, opened 2019-03-18
[GOVCMSD7-25] Password lockout - improve flood control in GovCMS 7

pandaskii:


gargsuchi (approved):
Looks good.

modules/custom/govcms_account_security/govcms_account_security.install
```diff
@@ -5,22 +5,41 @@
...
   // ISM 2015, control: 1403, accounts are locked after a maximum of 5 failed
   // logon attempts.
   variable_set('user_failed_login_user_limit', 5);
-  // Lock the offending account for 315400000 seconds (10 years).
-  variable_set('user_failed_login_user_window', 315400000);
+  // Lock the offending account for 21600 seconds (6 hours).
+  variable_set('user_failed_login_user_window', 21600);
   // Register flood events based on the uid only, see user.module.
   variable_set('user_failed_login_identifier_uid_only', TRUE);
-  // Set HTTP status code returned to anonymous visits on user profile pages.
```
fubarhouse, line 19: Could you please just verify this change is intentional?

modules/custom/govcms_account_security/govcms_account_security.install
```diff
@@ -41,31 +60,23 @@ function govcms_account_security_enable() {
...
-      'value' => $t('Enabled'),
-      'description' => $t('Login Security and Flood Control modules need to be disabled.'),
-      'severity' => REQUIREMENT_INFO,
-    );
-  }
-  return $requirements;
-}
-
-/**
- * Implements hook_uninstall().
```
fubarhouse, line 61: Why are we removing the uninstall process?
It would mean a user would have to manually override settings provided by the module that aren't documented if they were to detach from the product? I do not see this as ideal, please verify this decision.

modules/custom/govcms_account_security/govcms_account_security.install
```diff
@@ -41,31 +60,23 @@ function govcms_account_security_enable() {
...
-function govcms_account_security_uninstall() {
-  // Delete the variables to use the value set by Drupal's user module.
-  // Drupal's default value is 5.
-  variable_del('user_failed_login_user_limit');
-  // Drupal's default value is 21600.
-  variable_del('user_failed_login_user_window');
-  // Delete account security specific var.
-  variable_del('govcms_account_security_user_profile_page_code_anonymous_visit');
-}
+ function govcms_account_security_update_7000(&$sandbox) {
```
fubarhouse, line 65: Assuming there's the intent of not checking values before changing them?
Not a problem, just my opinion that this is a bit aggressive.

tests/behat/bootstrap/AccountContext.behat.inc
```diff
@@ -343,4 +343,100 @@ class AccountContext extends DrupalSubContextBase {
...
+   */
+  public function theFloodTableHasBeenCleared() {
+    $this->clearFloodTable();
+  }
+
+  /**
+   * Clear the flood table.
+   */
+  public function clearFloodTable() {
+    db_delete('flood')->execute();
```
fubarhouse: Would be much more comfortable with this being a call to `db_truncate`.
There are no conditions passed to this API call, and would otherwise delete the entire table, instead of truncating.

fubarhouse (approved):

PR #856, opened 2019-03-19
Update solr Dockerfile to be consistent with lagoon image

fubarhouse:
This PR will prevent PR test builds from inheriting D8 solr schema.xml on D7.


tobybellwood (approved):
We implemented the upstream 'pure' solr image in govCMS/govcmslagoon/pull/62 - this PR reflects that

PR #873, opened 2019-05-13
[GOVCMS-2972] Update services module version to 7.x-3.24 (from 7.x-3.23)

fubarhouse:
## Release notes
[Services - Less critical - Access bypass - SA-CONTRIB-2019-043](https://www.drupal.org/sa-contrib-2019-043)
Fixes: Services private file access bypass


pandaskii:
In favour of #868

PR #876, opened 2019-05-17
Quick fix the Behat tests issue

pandaskii:


tobybellwood (approved):
yup - sensible given the back and forth on these line endings!

PR #879, opened 2019-06-11
Remove GMP from the govCMS build.

steveworley:
GMP is only required for d7 if the openid module is being used.

- Removes GMP from cli image.
- Removes GMP from php image.

image


tobybellwood:
@steveworley -  this'll need to be against the govcmslagoon repo - we don't use the govcmsdev images any more :-)

PR #883, opened 2019-07-24
[GOVCMSD7-266] Patch SwiftMailer to fix the webform attachment issue

pandaskii:
## Problem/Motivation

Webform (7.x-4.19) feature "include files as attachments" is not functional with Swiftmailer

It is caused by a change in the webform module in the new release since 7.x-4.17:

SwiftMailer module requires the removed "$file['filepath']" to locate the attachment. Since this variable is removed, SwiftMailer takes it as a not valid attachment and the function is broken.

https://www.drupal.org/node/2921029

```
diff --git a/components/file.inc b/components/file.inc
index 47cf46f..cec9a5a 100644
--- a/components/file.inc
+++ b/components/file.inc
@@ -497,8 +497,6 @@ function _webform_delete_file($component, $value) {
  */
 function _webform_attachments_file($component, $value) {
   $file = (array) webform_get_file($value[0]);
-  // This is necessary until the next release of mimemail is out, see #1388786: mimemail_html_body fails to retrieve file's URI.
-  $file['filepath'] = $file['uri'];
   $files = array($file);
   return $files;
 }
```

## Proposed resolution

Apply the patch from Drupal.org

https://www.drupal.org/node/3069836


fubarhouse:
This was merged by 933d0a8949b59351919f5576c33530202575ad17

PR #886, opened 2019-08-09
[GOVCMSD7-208] Fixed undefined index notice.

sonnykt:
Fix the PHP notice
```
Notice: Undefined index: modules in govcms_account_security_form_alter() (line 90 of /app/profiles/govcms/modules/custom/govcms_account_security/govcms_account_security.module).
Warning: Invalid argument supplied for foreach() in govcms_account_security_form_alter() (line 90 of /app/profiles/govcms/modules/custom/govcms_account_security/govcms_account_security.module).
```
JIRA: https://jira.gold.gov.au/jira/browse/GOVCMSD7-208


mingsong-hu (approved):
Source code review passed. The error message should be gone with the new codes.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -87,6 +87,10 @@ function govcms_account_security_form_alter(&$form, &$form_state, $form_id) {
       // Disable the ability to change any configured modules.
       global $user;
       $modules = variable_get('govcms_account_security_policy_report_banned_modules', array());
+      if (empty($form['modules'])) {
```
pandaskii, line 86: Better to avoid a conditional break in a switch statement.

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -87,6 +87,10 @@ function govcms_account_security_form_alter(&$form, &$form_state, $form_id) {
...
+        break;
+      }
+
       foreach ($form['modules'] as $group => $value) {
```
pandaskii: Not sure foreach ((array)$form['modules'] as $group => $value) working or not

pandaskii (changes requested):

modules/custom/govcms_account_security/govcms_account_security.module
```diff
@@ -87,6 +87,10 @@ function govcms_account_security_form_alter(&$form, &$form_state, $form_id) {
       // Disable the ability to change any configured modules.
```
pandaskii: Expecting to see "if statement" here, thanks

For example

```
case 'system_modules':
   
   if (!check_this_variable_is_set_or_a_not_empty_array($form['modules'])) {
   ...
  }

break;
```

mingsong-hu: According to the official document, the PHP empty() function will check if a variable exists and if it is an empty array. Both cases, a TRUE value will be returned. Details see:
https://www.php.net/manual/en/function.empty.php

Therefore, I believe using the empty() function is good enough to avoid PHP warning message from generating.

modules/custom/govcms_account_security/govcms_account_security.module
sonnykt: when `$form['modules']` does not exist, `(array) $form['modules']` still raises a notice.

modules/custom/govcms_account_security/govcms_account_security.module
sonnykt: `break` is completely fine in this `case`.

modules/custom/govcms_account_security/govcms_account_security.module
simesy: Empty seems ok. I'd probably put it as the first thing inside the switch case as an "early return".

modules/custom/govcms_account_security/govcms_account_security.module
pandaskii: Yes, you are correct

modules/custom/govcms_account_security/govcms_account_security.module
pandaskii: This is a good point. thanks @simesy

pandaskii (changes requested):

pandaskii (approved):

pandaskii:
Thanks @sonnykt

PR #889, opened 2019-09-19
Updated README.md to include step to enable Pygmy

T-Fletcher:
Without knowing to first enable `pygmy`, users are left wondering why their govCMS site can't be loaded.


pandaskii (approved):
Good point, thanks Tim

PR #894, opened 2019-11-15
[GOVCMSD7-294] Upgrade PHP version to 7.2

mingsong-hu:
Upgrade PHP from 7.1 to 7.2

PR #895, opened 2019-11-20
[GOVCMS-3905] Use TugboatQA to build PR test environments for distro/modules/themes

mingsong-hu:


.tugboat/config.yml
```diff
@@ -0,0 +1,66 @@
+services:
+  app:
+    # Use PHP 7.2 with Apache to serve the Drupal site
+    image: tugboatqa/php:7.2-apache
```
pandaskii, line 4: `image: tugboatqa/php:7.2-apache`  - a little bit early in my mind. It will be good if #894 resolved

.tugboat/config.yml
```diff
@@ -0,0 +1,66 @@
...
+        - rm -rf "${DOCROOT}/sites/default/"
+        - mkdir "${DOCROOT}/sites/default/"
+        - mkdir -p "${DOCROOT}/sites/default/files"
+        - chgrp -R www-data "${DOCROOT}/sites/default/files"
+        - find "${DOCROOT}/sites/default/files" -type d -exec chmod 2775 {} \;
+        - find "${DOCROOT}/sites/default/files" -type f -exec chmod 0664 {} \;
+        - cp "${TUGBOAT_ROOT}/.tugboat/settings.php" "${TUGBOAT_ROOT}/docroot/sites/default/"
+       
+        # Install the site.
+        - drush -r "${DOCROOT}" si -y govcms
```
pandaskii, line 57:
```
# Install the fresh GovCMS site and do not send email notification
# - build/bin/phing -f build/phing/build.xml build:install
- drush -r "${DOCROOT}" site-install govcms install_configure_form.update_status_module='array(FALSE,FALSE)' -y
```

.tugboat/config.yml
```diff
@@ -0,0 +1,66 @@
...
+        - drush make --no-core drupal-org.make ${TUGBOAT_ROOT}/docroot/ -y
+       
+        # Prepare for install.
+        - rm -rf "${DOCROOT}/sites/default/"
+        - mkdir "${DOCROOT}/sites/default/"
+        - mkdir -p "${DOCROOT}/sites/default/files"
+        - chgrp -R www-data "${DOCROOT}/sites/default/files"
+        - find "${DOCROOT}/sites/default/files" -type d -exec chmod 2775 {} \;
+        - find "${DOCROOT}/sites/default/files" -type f -exec chmod 0664 {} \;
+        - cp "${TUGBOAT_ROOT}/.tugboat/settings.php" "${TUGBOAT_ROOT}/docroot/sites/default/"
```
pandaskii, line 54: It is OK to use ${TUGBOAT_ROOT}/docroot but ${DOCROOT} is better

.tugboat/config.yml
```diff
@@ -0,0 +1,66 @@
...
+        # Link the document root to the expected path. This example links
+        # /docroot to the docroot
+        - ln -snf "${TUGBOAT_ROOT}/docroot" "${DOCROOT}"
+
+      # Commands that import files, databases,  or other assets. When an
+      # existing preview is refreshed, the build workflow starts here,
+      # skipping the init step, because the results of that step will
+      # already be present.
+      update:
+        # We have no external assets to update for testing the distro.
```
pandaskii, line 38: `update` section is useful when `refreshing` the build workflow. The current implement is empty which may lead to a build error in the tugboat

.tugboat/config.yml
mingsong-hu: I think we need to test it with PHP 7.2

.tugboat/config.yml
mingsong-hu: Empty section is fine and tested. 
With GovCMS8 Tugboat version, we leave this section empty too.
https://github.com/govCMS/govCMS8/blob/1.x/.tugboat/config.yml

.tugboat/config.yml
mingsong-hu: No need to downgrade to PHP 7.1

.tugboat/config.yml
pandaskii: We need a fix to D8 then

PR #907, opened 2020-03-18
[GOVCMSD7-156] Remove pathauto_persist module

suhyeonh:
The module **pathauto_persist** has been merged into the Pathauto module hence no longer required  in the distribution.

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -104,7 +104,6 @@ projects[paranoia][version] = "1.7"
 projects[password_policy][version] = "1.16"
 projects[pathauto][version] = "1.3"
 ; @TODO remove this after pathauto update.
-projects[pathauto_persist][version] = "1.4"
 projects[pathologic][version] = "3.1"
 projects[pci_update][version] = "1.0"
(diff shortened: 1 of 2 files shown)


ruwanl (approved):

drupal-org.make
```diff
@@ -104,7 +104,6 @@ projects[paranoia][version] = "1.7"
 projects[password_policy][version] = "1.16"
 projects[pathauto][version] = "1.3"
 ; @TODO remove this after pathauto update.
```
pandaskii: And this line #106 is not needed anymore

pandaskii (changes requested):

pandaskii:
Also, We don't need this function while the modules are removed from the distribution

```
/**
 * Implements hook_paranoia_hide_modules().
 *
 * @TODO remove this after pathauto update.
 */
function govcms_paranoia_hide_modules() {
  return array(
    'pathauto_persist' => 'Other',
    'govcms_register' => 'govCMS',
  );
}
```

drupal-org.make
suhyeonh: the line 106 is removed.

PR #911, opened 2020-03-20
Update PHP to v7.3

fubarhouse:


mingsong-hu:
While building the Tugboat preview for this PR, there is an error said "Please reinstall the libzip distribution". 

Adding following line into .tugboat/config.yml line 22

apt-get install libzip-dev

The issue is fixed.

PR #915, opened 2020-03-25
[GOVCMSD7-314] Exclude file entity not in the public folder from the search index

pandaskii:
**Problem/Motivation**

The search index with the file entities can be set up in the search_api module.

File entities don't have published/unpublished state so they all get indexed regardless whether they are private or public files.

For example, a client has a webform. The search is not indexing Webform submissions. It is indexing File entities which are created when a file is uploaded to a webform submission (private path).

**Proposed resolution**

Wrote a data alteration plugin regarding this issue

Pls help for your kindly review and test so I can improve it, thanks


simesy:
I like the look of it. 

(Link to issue https://www.drupal.org/project/search_api/issues/2910359)

ruwanl (approved):

PR #916, opened 2020-03-25
[Govcmsd7 167] Local simpletest always gives “returned 0 (0 bytes)”

suhyeonh:
**Problem/Motivation**

Whenever a test wants to make an HTTP request with $this->drupalGet(), the test reports a fail saying.

**Proposed resolution**

Check value of `$_SERVER['HTTP_HOST']` - 'http://nginx:8080'


pandaskii (approved):
Looks good, thanks

PR #918, opened 2020-04-01
[GOVCMSD7-281] Add site builder permissions for login destination

fubarhouse:
Adds the permission 'administer login destination settings' to the site builder role by default as per detailed in GOVCMSD7-281.


mingsong-hu (approved):
Since the hook_install will only be called the first time when the GovCMS is installed, this PR will only affect a new GovCMS site. For an existing site, still need other approach to assign the new permission to the sitebuilder role.

PR #933, opened 2020-04-30
[GOVCMSD7-340]Update facetapi 7.x-1.6

suhyeonh:
# facetapi 7.x-1.6
## Release notes
Changes since 7.x-1.5 - fix php 7.3.x compatibility AND numerous other fixes and improvements
#3057285: Missing semicolons can cause aggregated and compressed JS to fail
#3037777: Allow conditionally making facet links nofollow
#2950212: "Prevent crawlers from following facet links" checkbox is not working in Facet Display settings
#3057283: Use #attached instead of drupal_add_js
#3048278: Relocate calls to func_get_args() to beginnings of functions. To avoid warnings on Upgrading to PHP 7
#2977195: Sort facet items based on language accent
#2373023: Undefined index: field api
#2863400: Add theme hook suggestions to item_list
#2915610: Don't call theme_link() directly
#2899295: Filter active values get lost if field alias start with another's field alias that is configured to be single value
#1661476: Add the extract() method to FacetapiQueryTypeInterface
#2404173: Stray urlencode() disables some blocks
#2327997: Show Fewer links not working
#2866117: Theme hook current_search_accessible_markup not found.
#2496511: Wrong classes added to facet links when altered
#2715259: Missing documentation for Sort's 'requitements'
#1942922: Use a separate theme function for the active items in the current search blocks
#2617028: Use correct theme function to render an active item in the Current Search Block
#2617028: Use correct theme function to render an active item in the Current Search Block
#2833007: Coding standard related issues in facetapi.module file.
#1616518: In current search show active search term, just like a facet
#2682671: Numeric Facets string sort, how do we get them to numeric sort
#2672524: Wrong possibility to check checkbox with 0 results
#2563117: Documentation issue with getActiveItems() method
#2311585: Regression: 'Show more/fewer' links are no longer translatable
#2393363: PHP Fatal error: Class name must be a valid object or a string in sites/all/modules/facetapi/plugins/facetapi/adapter.inc on line 1308
#2817941: Add supporting of the i18n for the overridden facet label
#1665164: Override Facet Label
#2311585: Regression: 'Show more/fewer' links are no longer translatable
#1482304: Add class for active list item
#2076187: Provide better DX wrt to block hash deltas and facet names
#2194423: Add an option to toggle whether the facet count is displayed
#2378693: Notice: Undefined offset: 1 in FacetapiAdapter->processActiveItems() line 312
#2545130: Allow the map callback to exclude items
#2382697: Typos and Spelling - D7
#2159883: Date facets not displayed when the configured granularity is larger than the calculated granularity
#2000626: Add a hook that allows altering of facet definitions at runtime
#2244933: Allow parent terms in a hierarchy to be enabled / disabled independent of their children
#2205489: Add search-path as a token
#2545130: Allow the map callback to exclude items
#2382697: Typos and Spelling - D7


mingsong-hu (approved):
No known issue with 7.x-1.6 which can’t be fixed either now or as required. Code review has shown no issues which should prevent the proceedings. The module indicates to work as designed/expected with no code quality concerns.

PR #949, opened 2020-05-13
add patch for libraries

suhyeonh:
Patch added for libraries to resolve Autoload issue.


mingsong-hu:
The autoload.php is downloaded into the library folder sites/all/libraries/php-encryption as expected.

Looks good to me.

PR #955, opened 2020-06-12
[GOVCMSD7-157] Remove session_cookie_lifetime module from distribution

mingsong-hu:


mingsong-hu:
Quick went through the source codes of this module.

It is very light weight module in which just implements the hook_init function and a configuration form. Other functions just provide the data required.

The uninstall hook will remove all configuration required by this module.

See https://git.drupalcode.org/project/session_cookie_lifetime/-/blob/7.x-1.x/session_cookie_lifetime.install

**Conclusion:**
It is safe to remove this module.

mingsong-hu:
While deploying this update to existing sites, we have to make sure running following drush commands before removing this module.

drush pm-uninstall session_cookie_lifetime
drush cc all

simesy:
Good work AMDS

ruwanl (approved):

PR #988, opened 2021-05-26
Patching Views module to be compatible with PHP 7.4

ruwanl:
Patching Views module to be compatible with PHP 7.4


mingsong-hu (approved):
Looks good.

PR #989, opened 2021-05-26
[GOVCMSD7-408] Patch Paragraphs module to be compatible with PHP 7.4

ruwanl:
Patch Paragraphs module to be compatible with PHP 7.4


Gumnutt:
Looks good the patch is added and looks correct

Gumnutt (approved):

PR #990, opened 2021-05-26
[GOVCMSD7-409] Patch Entity Cache module to be compatible with PHP 7.4

ruwanl:
Patch Entity Cache module to be compatible with PHP 7.4


Gumnutt:
Looks good the patch is added and looks correct

Gumnutt (approved):

PR #998, opened 2021-05-28
Use PHP 7.4 with Apache to serve the test site

pandaskii:


mingsong-hu (approved):
PHP version and Drush version are updated as expected.
Good job.

